PR #58, opened 2014-06-03
Multiple Salesforce databases support

hynekcer:
Supported more Salesforce databases in an application.
- The setting SALESFORCE_DB_ALIAS is the name of the primary Salesforce
    database. It is possible to completely change the precedence of databases
    and 'salesforce' can be the name of secondary SF database.
- The SF database can be selected by `.objects.using('some_sf').filter()`
    and by `.save(using='...')` for saving new objects.
- Foreign keys correctly relate to objects in the same database.
- An obsoleted auth session causes reauthentification of the right database.
- Some some features can be tested only if more Salesforce databases are
    defined in DATABASES settings.
- Modified auth.authenticate parameters so that the old usage should raise
    a clear exception. (though a compatible solution was functional, but it was
    less readable.)

It is implemented in the first commit.

More commits are added with different small features and fixes.

### DIFF
--- a/salesforce/backend/compiler.py
+++ b/salesforce/backend/compiler.py
@@ -15,17 +15,6 @@ from django.db.models.sql.datastructures import EmptyResultSet
 from salesforce import DJANGO_15, DJANGO_16, DJANGO_17_PLUS
 
 
-def process_name(name):
-	"""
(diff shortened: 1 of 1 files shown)


Emma8684 (commented):
Gcjjtfbfjk

Issue #83, opened 2014-10-24
What is the 'HOST' of DATABASES={'salesforce':{...'HOST':''}} within settings.py?


TaoPercolate:
I'm trying to use django-salesforce, but I still cannot pass authorization. I'm wondering if it's because I don't set 'HOST' right. 

My error message is:

LookupError: oauth failed: email@my.com: {"error_description":"expired access/refresh token","error":"invalid_grant"}

PR #91, opened 2015-01-12
Expanding README with auth info

jamescooke:
Thanks for the great Django module.

I've struggled today getting the authorisation to work with Salesforce, primarily because I'm not familiar with their API and docs. I found all the info I needed in this closed issue: https://github.com/django-salesforce/django-salesforce/issues/83

As per the suggestion there, I've added this info to the quick start in README. Hopefully it'll help other Djangonauts who don't know Salesforce well in the future.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -53,7 +53,7 @@ Quick Start
    * ``USER`` is the username used to connect.
    * ``PASSWORD`` is a concatenation of the user's password and security token.
      Security token can be omitted if the local IP address has been
-     whitelisted.
+     whitelisted in Security Controls / Network Access.
    * ``HOST`` is ``https://test.salesforce.com`` to access the sandbox, or
      ``https://login.salesforce.com`` to access production.
 
--- a/README.rst
+++ b/README.rst
@@ -43,15 +43,32 @@ Quick Start
 
    * ``CONSUMER_KEY`` and ``CONSUMER_SECRET`` values are for the app used to
      connect to your Salesforce account. Instructions for how get these are in
-     the Salesforce REST API Documentation.
+     the Salesforce REST API Documentation. Key and secret can be created on
+     web by:
+     - Salesforce web > Setup > App Setup > Create > Apps > Connected apps >
+       New.
+     - Click "Enable OAuth Settings" in API, then select "Access and manage
+       your data (api)" from available OAuth Scopes.
+     - Other red marked fields must be filled, but are not relevant for Django.
    * ``USER`` is the username used to connect.
    * ``PASSWORD`` is a concatenation of the user's password and security token.
+     Security token can be omitted if the local IP address has been
+     whitelisted.
    * ``HOST`` is ``https://test.salesforce.com`` to access the sandbox, or
      ``https://login.salesforce.com`` to access production.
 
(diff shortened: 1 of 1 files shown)


hynekcer:
Yes. It is not all. For `CONSUMER_KEY` and `CONSUMER_SECRET` can be written

> You can create them on the web by 
> ` ` ` Setup / App Setup / Create / Apps / Connected apps / New`. 
> Fill any names into `Basic information`. 
> Click `Enable OAuth Settings` in the section `API (Enable OAuth Settings)`. 
> Fill any unimportant `Callback URL`. 
> Select some OAuth Scopes from `Available OAuth Scopes`. 
> Save

The minimum is probably "Access and manage your data (api)".
I usually select only "Full access (full)", because I usually want to get the concievable maximum including the access to metadata and I have not tried anything less yet.

What is your latest experience?

philchristensen:
Glad to accept this patch, but it would be great if you could clarify that the security code can be omitted from the password if the local IP address has been whitelisted (although certainly for new users, it may be worth circumventing the IP whitelist with the security token).

jamescooke:
@hynekcer That looks like good additional info. By searching for some of the phrases you've used, I've found this page: https://help.salesforce.com/apex/HTViewHelpDoc?id=connected_app_create.htm , which has very detailed information. Would it be good to link to this from the README rather than duplicating Salesforce info?

My experience yesterday was that I was given access credentials for the sandbox system, but I hadn't realised that the password and token should be concatenated.

@philchristensen Thanks for the feedback. Happy to add that information about the possible omission of the security code - is there a link which the README could point people at for more info?

hynekcer:
Yes, the link is very useful, otherwise there would be questions permanently. But it is too long, complicated and not all.

crib under the desk:
> Salesforce web / Setup / App Setup / Create / Apps / Connected apps / New. 
> Click "Enable OAuth Settings" in API, then select "Access and manage your data (api)" from available OAuth Scopes.
> Other red marked fields must be filled, but are not relevant for Django.

All test passed with this, without full access. Even Tooling API works good.

Everything for a project can work under restricted Salesforce user account if it has access to objects in your models. Introspection (inspectdb) doesn't require any permissions. Running tests for django_salesforce requires many permissions or Administrator account for sandbox.

jamescooke:
@hynekcer Good information and crib sheet :+1: I've added this info.

@philchristensen I've added the info about whitelisting.

hynekcer:
That whitelisting is really great.

I prefer to add details:

  ... if the local IP address has been whitelisted 
+ in Security Controls / Network Access.

(I would like to add some authentication method without password and it will be able to run with minimal requirements one day.)

jamescooke:
@hynekcer Cool. I've added that.

As I get more familiar with this module I can see that there's some space for some documentation. Had you ( @hynekcer @philchristensen ) got anywhere with that?

philchristensen:
Thanks for the README fixes @jamescooke ; patches for any other documentation are always welcome!

hynekcer:
Yes, thanks. I'm fully occupied by writing code that I can write documentation only for entirely new features. Then I must ask someone someone here to fix my English. I can answer some question to you or tell what is in development and that is better to prepare that text for the new version.

Issue #43, opened 2014-05-01
Impossible to resolve relationship caused by several driving SObject


pirhoo:
Hi,

the following code throws me a curious error from SalesForce :

``` python
# Get all voyages for the given account
voyages = DemandeDeFinancement.objects.all().filter(Account__rne='0930897V')
```

Results on :

> A driving SObject type has already been set, all other entity types in the FROM clause must be relationships to the initial object.  The driving object is Demande_de_financement__c

Any clue? 

This library is still amazing :)

Best,
Pierre

hynekcer:
The issue is fixed now in the latest master. (it should work with your original filter)

PR #92, opened 2015-01-23
Allow to select individual fields of related object

GaretJax:
This PR allows simple querying (retrieving and filtering) on related objects by mapping just some fields as normal fields on the parent model.

Consider the following model, with a custom relationship with `Opportunity`

``` python
class SupplierOpportunity(models.Model):
    name = models.CharField(max_length=120, db_column='Name')
    # ... other fields ...
    stage_name = models.CharField(max_length=150, db_column='Opportunity__r.StageName')

    class Meta:
        db_table = 'Opportunity_Supplier__c'

```

It is now possible to query this model and access the related attributes as its own fields:

``` python
# Filtering
obj_in_stage = SupplierOpportunity.objects.filter(stage_name='...')

# Accessing
stage_name = SupplierOpportunity.objects.get(pk='...').stage_name
```

P.S.: I am a total n00b with salesforce, but just happened to solve a filtering performance issue this way. If there are better ways to implement this, please don't hesitate to point it out.

### DIFF
--- a/salesforce/backend/query.py
+++ b/salesforce/backend/query.py
@@ -174,7 +174,9 @@ def prep_for_deserialize(model, record, using, init_list=None):
 				fields[x.name] = record[simple_column]['Type']
 			else:
 				# Normal fields
-				field_val = record[x.column]
+				field_val = record
+				for col in x.column.split('.'):
+					field_val = field_val[col]
 				#db_type = x.db_type(connection=connections[using])
 				if(x.__class__.__name__ == 'DateTimeField' and field_val is not None):
 					d = datetime.datetime.strptime(field_val, SALESFORCE_DATETIME_FORMAT)


philchristensen:
Thanks for the contribution. I'm not sure how I feel about having fields from another model appear as part of a different one. I'd like the syntax to reflect the fact that you're merely retrieving the value, and not able to change it. Maybe it'd end up like `<SalesforceQuerySet>.related_value('stage_name')` for retrieval, with the queryset supporting normal filtering syntax, e.g., `filter(opportunity__StageName='...')`

I'd like to get @hynekcer's input on things as well.

GaretJax:
I tend to agree with you. I like your proposition, makes it more explicit.

We have `sf_read_only=models.READ_ONLY` for all our fields anyway... as we're only reading them. Maybe, instead of declaring a normal field, we can use some sort of notation to make it clear:

    stage_name = RelatedField('Opportunity__r', 'StageName')

(just some thoughts... we would still have to define a type somewhere though)

hynekcer:
Lookup (filtering) by fields of related object is easy.

    my_queryset = SupplierOpportunity.objects.filter(opportunity__stage_name='...')

I think all combinations of custom fields, builtin fields, custom objects and builtin objects are supported. It is compiled by one query. You can investigate the SOQL of the query set by `my_queryset.query.get_compiler('salesforce').as_sql()` and see, that it is compiled by one query.

I rename the issue therefore.

Yes, access to a field of the related object is a nice hack. The *patch* can not affect the rest of the package. I am afraid that the hacked model can not be generally recommended in the documentation.

It works for *Master relationship detail* fields, but it raises TypeError for empty *Lookup* fields. Please FIX it.

Yes, you use READ_ONLY and if you omit it than the hacked object can not be created or updated.
```
stage_name = models.CharField(sf_read_only=models.READ_ONLY`, db_column='Opportunity__r.StageName', max_length=150)
```
I should write model validation that recects an incorrect field with dot and without read only.

I was afraid of [governor limits](https://www.salesforce.com/us/developer/docs/apexcode/Content/apex_gov_limits.htm) that some limit would be divided by two if twice more rows must be accessed but I currently don't see any difference in size or speed. I can access more than 40000 records in 20 seconds. The whole number of fields is very important for speed and can be restricted by `only(...)` method, but one related field is not essential for it at all.

I was working on a solution with `select_related` or `prefetch_related`, but it is really very complicated and adding many field is expensive in Salesforce.

I am inspired by your solution and I am thinking how a detail of related object can be added dynamically without hacking the model.

hynekcer:
@philchristensen It is a great idea with `select_related`. It does not check the type of field too early and it is possible to get fields that are not foreign keys. Example

    >>> qs = SupplierOpportunity.objects.filter(opportunity__stage_name='...'
    ...             ).select_related('opportunity__stage_name')
    >>> print qs.query.select_related
    {'opportunity': {'stage_name': {}}}

It is not perfect but I can get this information in salesforce.backend.compiler, remove such "bugs" before passing it into Django qury compiler and to add only exactly that "incorrect" fields to compiler results. The first step will be that such fields will be accessed by dict

    for item in qs:
        print item.my_raw_dict['opportunity']['stage_name']

...later by a dot. I am continuing.

hynekcer:
Excuse, I tested your issue on a branch that is not yet committed to master. Lookup is still compliucated for your. The fix fill work only on Django 1.7 and newer. Is it acceptable?

GaretJax:
@hynekcer are you talking about the PR? I just used it on Django 1.5.5 and it works, while I didn't even try the plain `filter`.

I think my problem stems from the fact that I already have another field named opportunity (I think it's because of "polymorphic" relationships... but as I said, I have no clue and I just inherited this code):

```python
class SupplierOpportunity(models.Model):
    name = models.CharField(max_length=120, verbose_name='Name', db_column='Name', sf_read_only=models.READ_ONLY)
    stage_name = models.CharField(max_length=150, db_column='Opportunity__r.StageName', sf_read_only=models.READ_ONLY)
    opportunity = models.ForeignKey(
        Account, db_column='Opportunity__c', sf_read_only=models.READ_ONLY, on_delete=models.DO_NOTHING)
```

so filtering on `opportunity__stage_name` will not work I guess.

hynekcer:
A similar issue is #43 that hasn't been yet fixed completely, but also evaded by the user. You know, that a big problem is in Salesforce [SOQL](http://www.salesforce.com/us/developer/docs/soql_sosl/). It uses the suffix `__r` for access to fieds of related object, while suffix `__c` or no suffix for access only to key of this object. I think it is not a result of polymorthy, but of their previous catchword "no software". :-) My [fix that correctly translates suffixes](https://github.com/hynekcer/django-salesforce/commit/f8a03293422030c6130d61fab976475abd5541cb) between Django QueryCompiler and Salesforce QueryCompiler exist since end of October but had not been committed due to some missing tests and uncleaned code. It should work correctly also with `get` in Django 1.7+.

OK. In Django 1.5.x you will have to use your solution with the hacked model and I am not breaking it by rewriting.

GaretJax:
Thanks for the clarification @hynekcer!

hynekcer:
Individual fields of related objects can be now added by standard Django methods like `values`, `values_list` or a complete related object by method `select_related("related_name")` and many others.

PR #93, opened 2015-01-24
Fixed compatibility with Posgresql, interited AutoField, closed #90

### LINKED ISSUE #90: SalesforceAutoField does not inherit from fields.AutoField
I'm running into an error running tests on a postgres (non-Salesforce) database. The issue is that django doesn't respect `SalesforceAutoField.get_internal_type`, and generates errors like the following:
```
IntegrityError: null value in column "Id" violates not-null constraint
```

Django is not properly viewing the field as an auto-increment primary key. I think this is caused by a bug in Django [here](https://github.com/django/django/blob/01ab84c61330ffa5ac87c637249611c5e5343e57/django/db/models/base.py#L800), where it checks whether the field is an instance of `AutoField` instead of checking that `f.get_internal_type() == 'AutoField'`. The Django source code seems to be inconsistent on this point, sometimes checking `get_internal_type`, and sometimes checking whether the field is an instance of `AutoField`. I'm going to submit a patch with test cases to Django, but an easy fix in `django-salesforce` itself would be to have `SalesforceAutofield` inherit from `fields.AutoField`.

## How to reproduce
Set your settings file to use a postgres db:
```python
DATABASES = {
    'default': {
        'NAME': 'something',
        'ENGINE': 'django.db.backends.postgresql_psycopg2',
        'USER': 'django',
        'PASSWORD': 'secret',
    },
}
```
Define a simple model:
```python
from salesforce import models as sf_models
from django.db import models

class Foo(sf_models.Model):
    class Meta:
        managed=True
```

Run:
 ```
$ DJANGO_SETTINGS_MODULE=django_salesforce_example python manage.py syncdb
$ DJANGO_SETTINGS_MODULE=django_salesforce_example python manage.py shell
```
Then run in the python shell run
```python
>>> from django_salesforce_example.models import *
>>> Foo.objects.create()
```
This generates the following stack trace:
```pytb
  File "<console>", line 1, in <module>
  File "/Users/lucaswiman/scratchpads/work/django_salesforce_example/.venv/default/lib/python2.7/site-packages/django/db/models/manager.py", line 157, in create
    return self.get_queryset().create(**kwargs)
  File "/Users/lucaswiman/scratchpads/work/django_salesforce_example/.venv/default/lib/python2.7/site-packages/django/db/models/query.py", line 322, in create
    obj.save(force_insert=True, using=self.db)
  File "/Users/lucaswiman/scratchpads/work/django_salesforce_example/.venv/default/lib/python2.7/site-packages/django/db/models/base.py", line 545, in save
    force_update=force_update, update_fields=update_fields)
  File "/Users/lucaswiman/scratchpads/work/django_salesforce_example/.venv/default/lib/python2.7/site-packages/django/db/models/base.py", line 573, in save_base
    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
  File "/Users/lucaswiman/scratchpads/work/django_salesforce_example/.venv/default/lib/python2.7/site-packages/django/db/models/base.py", line 654, in _save_table
    result = self._do_insert(cls._base_manager, using, fields, update_pk, raw)
  File "/Users/lucaswiman/scratchpads/work/django_salesforce_example/.venv/default/lib/python2.7/site-packages/django/db/models/base.py", line 687, in _do_insert
    using=using, raw=raw)
  File "/Users/lucaswiman/scratchpads/work/django_salesforce_example/.venv/default/lib/python2.7/site-packages/django/db/models/manager.py", line 232, in _insert
    return insert_query(self.model, objs, fields, **kwargs)
  File "/Users/lucaswiman/scratchpads/work/django_salesforce_example/.venv/default/lib/python2.7/site-packages/django/db/models/query.py", line 1514, in insert_query
    return query.get_compiler(using=using).execute_sql(return_id)
  File "/Users/lucaswiman/scratchpads/work/django_salesforce_example/.venv/default/lib/python2.7/site-packages/django/db/models/sql/compiler.py", line 903, in execute_sql
    cursor.execute(sql, params)
  File "/Users/lucaswiman/scratchpads/work/django_salesforce_example/.venv/default/lib/python2.7/site-packages/django/db/backends/util.py", line 69, in execute
    return super(CursorDebugWrapper, self).execute(sql, params)
  File "/Users/lucaswiman/scratchpads/work/django_salesforce_example/.venv/default/lib/python2.7/site-packages/django/db/backends/util.py", line 53, in execute
    return self.cursor.execute(sql, params)
  File "/Users/lucaswiman/scratchpads/work/django_salesforce_example/.venv/default/lib/python2.7/site-packages/django/db/utils.py", line 99, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "/Users/lucaswiman/scratchpads/work/django_salesforce_example/.venv/default/lib/python2.7/site-packages/django/db/backends/util.py", line 53, in execute
    return self.cursor.execute(sql, params)
IntegrityError: null value in column "Id" violates not-null constraint
DETAIL:  Failing row contains (null).
```

hynekcer:
I think that it also simplifies the future development.

### DIFF
--- a/coverage_run.sh
+++ b/coverage_run.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 # expects that "tox" has been run
 .tox/py27dj17/bin/coverage run --source salesforce --omit 'salesforce/packages/*' manage.py validate >/dev/null
-for x in py26dj14 py27dj17 py33dj15 py34dj16; do
+for x in py26dj14 py27dj15 py33dj17 py34dj16; do
     .tox/${x}/bin/coverage run -a --source salesforce --omit 'salesforce/packages/*' manage.py inspectdb --database=salesforce >/dev/null
     .tox/${x}/bin/coverage run -a --source salesforce --omit 'salesforce/packages/*' manage.py test salesforce
 done
--- a/salesforce/fields.py
+++ b/salesforce/fields.py
@@ -41,7 +41,7 @@ if not SF_PK in ('id', 'Id'):
 	raise ImproperlyConfigured("Value of settings.SF_PK must be 'id' or 'Id' or undefined.")
 
 
-class SalesforceAutoField(fields.Field):
+class SalesforceAutoField(fields.AutoField):
 	"""
 	An AutoField that works with Salesforce primary keys.
 	"""
@@ -53,7 +53,7 @@ class SalesforceAutoField(fields.Field):
 	}
 	def __init__(self, *args, **kwargs):
 		assert kwargs.get('primary_key', False) is True, "%ss must have primary_key=True." % self.__class__.__name__
-		kwargs['blank'] = False
+		kwargs['blank'] = True
 		kwargs['null'] = False
 		kwargs['default'] = None
 		fields.Field.__init__(self, *args, **kwargs)
--- a/salesforce/fields.py
+++ b/salesforce/fields.py
@@ -44,34 +44,24 @@ if not SF_PK in ('id', 'Id'):
 class SalesforceAutoField(fields.AutoField):
 	"""
 	An AutoField that works with Salesforce primary keys.
+
+	It is used by SalesforceModel as a custom primary key. It doesn't convert
+	its value to int.
 	"""
 	description = _("Text")
-	
-	empty_strings_allowed = True
+
 	default_error_messages = {
 		'invalid': _('This value must be a valid Salesforce ID.'),
 	}
-	def __init__(self, *args, **kwargs):
-		assert kwargs.get('primary_key', False) is True, "%ss must have primary_key=True." % self.__class__.__name__
-		kwargs['blank'] = True
-		kwargs['null'] = False
-		kwargs['default'] = None
-		fields.Field.__init__(self, *args, **kwargs)
-	
-	def get_internal_type(self):
-		return "AutoField"
-	
+
 	def to_python(self, value):
 		if isinstance(value, string_types) or value is None:
 			return value
 		return smart_text(value)
-	
-	def validate(self, value, model_instance):
-		pass
-	
+
 	def get_prep_value(self, value):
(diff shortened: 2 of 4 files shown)


philchristensen:
@hynekcer: Seems like we should do a point-release for something like this. Anything on the horizon I should wait for?

Incidentally, I'll soon be back to using django-salesforce again fairly heavily at my main job. I'm encouraging the sales/marketing deparments to convert fully to Salesforce from a home-grown CRM specifically so I can have a standard interface to interact with their data ;-)

hynekcer:
I have currently 24 commits not added to any pull request, waiting for final testing, and 10 commits waiting in the queue for finalizing. The most important open feature is the fix for ForeignKey filter (issue 43) from the end of October. It was relative easy but it is complicated to write tests

I'm reoganizing the majority of commits listed in the release draft. Sometimes is the order important for better testing, sometimes I want to do folow-up steps of one feature in a successive group of commits even though I developed them like multiplexed by othe features, sometimes I changed anything in the example model and tests must be updated. Once I found a bug much later that a test did not tested anything. I try to follow the standard that every commit in PR works as expected or a known bug is described in the commit message that will be fixed exactly in the next commit, if it is better to separate them

E.g. the SFDC API 32.0 caused a bug, because they added an incorrect ForeignKey that is not related to any model (undocumented and probably also unused). So I first demonstrated the bug by increasing the API, then I fixed inspectdb and finally I fixed also the guaranted validity of the exported Django model. (by automatic renaming such ForeignKey to TextField and writing a big comment behind it)

philchristensen:
So you found a bug in the Salesforce v32.0 API because of the django-salesforce unit tests? That's great!

I'm glad to do a release whenever. At this point it seems like mostly edge cases that have been fixed since the last release.

-phil

PR #96, opened 2015-02-10
Integration of convertLead() from the SOAP API using beatbox

tvanesse:
A new utility module that adds support for the Salesforce SOAP API in order to take advantage of the missing endpoints in the REST API (such as convertLead).
This feature requires an additional dependency as we use a third-party module to talk to the SOAP API: [beatbox](https://pypi.python.org/pypi/beatbox/32.1). However, this dependency is totally optional for users who do not need more than the capabilities offered by the REST API.

Launch the unit test with

```
./manage.py test salesforce.tests.test_utils
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ coverage_html
 log
 tests/inspectdb/models.py
 salesforce/packages/
+venv
--- a/README.rst
+++ b/README.rst
@@ -231,18 +231,18 @@ Advanced usage
      lead = Lead.objects.all()[0]
      response = convert_lead(lead)
 
-    For the particular case of ``Lead`` conversion, beware that having
-    some *custom* and *required* fields in either ``Contact``,
-    ``Account`` or ``Opportunity`` is not supported. This arises from
-    the fact that the conversion mechanism on the Salesforce side is only
-    meant to deal with standard Salesforce fields, so it does not really
-    care about populating custom fields at insert time.
+   For the particular case of ``Lead`` conversion, beware that having
+   some *custom* and *required* fields in either ``Contact``,
+   ``Account`` or ``Opportunity`` is not supported. This arises from
+   the fact that the conversion mechanism on the Salesforce side is only
+   meant to deal with standard Salesforce fields, so it does not really
+   care about populating custom fields at insert time.
 
-    One workaround is to map a custom required field in
-    your `Lead` object to every custom required field in the target
-    objects (i.e., `Contact`, `Opportunity` or `Account`). Follow the
-    `instructions<http://www.python.org/https://help.salesforce.com/apex/HTViewHelpDoc?id=customize_mapleads.htm>`_
-    for more details.
+   One workaround is to map a custom required field in
+   your `Lead` object to every custom required field in the target
+   objects (i.e., `Contact`, `Opportunity` or `Account`). Follow the
+   `instructions <http://www.python.org/https://help.salesforce.com/apex/HTViewHelpDoc?id=customize_mapleads.htm>`__
+   for more details.
 
 
 
--- a/README.rst
+++ b/README.rst
@@ -216,6 +216,35 @@ Advanced usage
    name ``Contact`` and all tables that are prefixed with ``Account``. This
    filter works with all supported database types.
 
+-  **Accessing the Salesforce SOAP API** - There are some Salesforce actions that cannot or can hardly
+   be implemented using the generic relational database abstraction and the REST API.
+   For some of these actions there is an available endpoint in the old Salesforce API
+   (SOAP) that can be accessed using our utility module. In order to use that module,
+   you will need to install an additional dependency ::
+
+     pip install beatbox
+
+   Here is an example of usage with ``Lead`` conversion ::
+
+     from salesforce.utils import convert_lead
+
+     lead = Lead.objects.all()[0]
+     response = convert_lead(lead)
+
+    For the particular case of ``Lead`` conversion, beware that having
+    some *custom* and *required* fields in either ``Contact``,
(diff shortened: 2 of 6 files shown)


philchristensen:
This looks pretty good so far. One thing I'd like to see is rather than including beatbox in the requirements.txt file (which would force it to be installed for everybody) is to use an import exception hook to display a sensible error, e.g.:

```
try:
    import beatbox
except ImportError:
    log.error("To use convertLead, you'll need to install the Beatbox library.")
```

hynekcer:
Thanks, this is useful. How did you solved required custom fields? Did you find anything with SOAP API / beatbox or did you wrote a custom trigger for Lead conversion or anythin else?

I would solve the dependency on beatbox similarly like Django recommends to use `pytz` package but doesn't require it:

    try:
        import beatbox
        if beatbox._beatbox.__version__ < '32.0':
            beatbox = None
    except ImportError:
        beatbox = None
    ....
    def require_beatbox():
        if not beatbox
            raise Exception(.... clear text)
    ...
    # tests
    from django.utils.unittest import skipUnless
    ....
        @skipUnless(beatbox)
        def test_lead_conversion(self):
            ....

hynekcer:
@tvanesse Please, also delete the objects after tests, because this is Salesforce db.

tvanesse:
I was just asking myself how to deal with optional requirements but your proposals are quite neat.

As for the required custom fields there are several options but I did not take the time to deeply investigate. I think the way to go is to use what they call [field mapping](https://help.salesforce.com/apex/HTViewHelpDoc?id=customize_mapleads.htm&language=en). It basically allows you to say "this custom required field in `Contact` should be populated from this other field in `Lead` when it is converted". Also it seems this step is mandatory even if you don't use the API because the problem occurs whenever you try to convert a Lead (even from the web interface). So my understanding is that this "problem" does not fall in `django-salesforce` responsibilities but rather to the end user. What do you guys think?

hynekcer:
I'm reading that required fields are not required at Lead conversion, because it must not block conversions, but values must be provided at the first edit. (http://salesforce.stackexchange.com/a/36889) Please, can you verify this answer? Only verify it once on a temporary required custom field that a conversion is possible or not. (It can not be tested automatically on default models because destnation required fields are hardcoded mapped from required Lead fields.) If the conversion is possible, than a pessimistic TODO can be rewritten to a more optimistic README that an additional field mapping can be easily done by the user code in Django.

tvanesse:
I am a bit confused by the response provided on stackexchange because I actually **experienced** the problem with one of my own custom required fields on `Contact`. The API call sent a response that explicitly said it was impossible to insert a new entry because this specific field was custom and required, and the conversion did not succeed. Removing the requireness of this field solved the problem and led to a successful conversion.

Now since the guy scored more than 44k on stackexchange, it seems more reasonable to assume that I made a mistake somewhere :smile:  I will investigate further tomorrow. Thanks for your inputs.

tvanesse:
Ok here is the thing:

It is **not** possible to perform a `convertLead` if there exist at least one **custom, required** field in one of the target objects (*i.e.* `Contact`, `Account` or `Opportunity`). It was maybe possible to do that in previous versions of SF but not anymore. Some evidences:

custom_required_fail 
(the message you get when you try to set a custom field as required in `Contact`)

and the error message from `beatbox`:

> System.DmlException: Insert failed. First exception on row 0; first error: REQUIRED_FIELD_MISSING, Required fields are missing: [My_Field__c]: [My_Field__c] (System Code)UNKNOWN_EXCEPTION

Two solutions:

1. Map custom Lead fields to the custom required fields on the target objects.
    * Then you need to create a custom required field in Lead for every custom required fields in the target objects. Should not be a problem in my opinion.
2. Define a default value for this custom required field in case no value is provided. But then we can legitimately wonder why this field is "required".

In both scenarios I believe `django-salesforce` could at least warn the user regarding this issue, both in the comments and in the logs. Any thought?

hynekcer:
I think that the answer could be about APEX. I remember that some example APEX code of Lead conversion does it in two steps. An optimistic conversion step, manipulation with the new created objects like if they were converted and the final step which checks which Lead conversion succeeded. APEX code can use transactions inside a request and no network activity that waits for feedback is possible inside request, Every API request is done in one isolated transaction. The integrity is enforced so frequently.

A have read that the custom field mapping is used by the interactive web Lead conversion tool and it is not used elsewhere.

This simple solution is definitely useful.

tvanesse:
> A have read that the custom field mapping is used by the interactive web Lead conversion tool and it is not used elsewhere.

Come on, do they really understand the concept of API at SF? I mean a Lead conversion should be performed in exactly the same way regardless of  the interface (web UI, Apex or API).

Anyway, I don't know anything about Apex so I'm afraid I won't be of any help if this is the only alternative we have.

philchristensen:
So, what's the functionality now? Is it just that it won't work on Contacts with required custom fields?

tvanesse:
Yes exactly. For the time being it won't work if you have at least one *custom* and *required* field in either `Contact`, `Opportunity` or `Account`.

philchristensen:
That seems acceptable to me. This certainly feels like an SFDC problem ;-)

I'm good to merge this, but I want to make sure the Travis build is running the extra unit tests, so it will probably be later this weekend.

hynekcer:
Thomas, I only braked the optimistic conclusion, but I can not negate it. (I realized I believed contradictory statements that can be googled, though many look really serious and independent.)

The code seems very good now. I'm afraid about the documentation that nobody of us have tried the mapping personally and also that I see there my opinions.
 
Maybe the text can be added again if it could be verified experimentally. It would useful even with an optional source field in Lead and a logic can be build around in Python..

PR #103, opened 2015-05-14
Fix table name mapping on nested queries

Y3K:
Django 1.7 uses table name aliases on nested queries, here's a comparison of the same SQL queries against 1.6

```
Django 1.6
SELECT OpportunityLineItem.Id, OpportunityLineItem.OpportunityId FROM OpportunityLineItem WHERE (OpportunityLineItem.HasSchedule = %s  AND OpportunityLineItem.OpportunityId IN (SELECT Opportunity.Id FROM Opportunity WHERE (Opportunity.StageName = %s  AND Opportunity.Probability > %s  AND (Opportunity.CloseDate >= %s  OR Opportunity.Name LIKE %s ) AND Opportunity.Amount > %s )))

Django 1.7
SELECT OpportunityLineItem.Id, OpportunityLineItem.OpportunityId FROM OpportunityLineItem WHERE (OpportunityLineItem.HasSchedule = %s AND OpportunityLineItem.OpportunityId IN (SELECT U0.Id FROM Opportunity WHERE (U0.StageName = %s AND U0.Probability > %s AND (U0.CloseDate >= %s OR U0.Name LIKE %s) AND U0.Amount > %s)))
```

Notice how `Opportunity` gets replaced with `U0`, this modifies the API call as well, here's the comparison:
```
Django 1.6
query?q=SELECT+OpportunityLineItem.Id%2C+OpportunityLineItem.OpportunityId+FROM+OpportunityLineItem+WHERE+%28OpportunityLineItem.HasSchedule+%3D+true++AND+OpportunityLineItem.OpportunityId+IN+%28SELECT+Opportunity.Id+FROM+Opportunity+WHERE+%28Opportunity.StageName+%3D+%27Closed+Won%27++AND+Opportunity.Probability+%3E+70.0++AND+%28Opportunity.CloseDate+%3E%3D+2013-01-01++OR+Opportunity.Name+LIKE+%27Boost%2FOther%25%27+%29+AND+Opportunity.Amount+%3E+0.0+%29%29%29

Django 1.7
query?q=SELECT+OpportunityLineItem.Id%2C+OpportunityLineItem.OpportunityId+FROM+OpportunityLineItem+WHERE+%28OpportunityLineItem.HasSchedule+%3D+true+AND+OpportunityLineItem.OpportunityId+IN+%28SELECT+U0.Id+FROM+Opportunity+WHERE+%28U0.StageName+%3D+%27Closed+Won%27+AND+U0.Probability+%3E+70.0+AND+%28U0.CloseDate+%3E%3D+2013-01-01+OR+U0.Name+LIKE+%27Boost%2FOther%25%27%29+AND+U0.Amount+%3E+0.0%29%29%29
```

Raising the `SalesforceError`
```
SalesforceError: 
OpportunityLineItem.OpportunityId IN (SELECT U0.Id FROM Opportunity WHERE (U0.StageName
                                             ^
ERROR at Row:1:Column:186
Didn't understand relationship 'U0' in field path. If you are attempting to use a custom relationship, be sure to append the '__r' after the custom relationship name. Please reference your WSDL or the describe call for the appropriate names.
```

On this PR I'm demapping the table names in order to perform the right API calls.

### DIFF
--- a/salesforce/backend/query.py
+++ b/salesforce/backend/query.py
@@ -49,6 +49,21 @@ SALESFORCE_DATETIME_FORMAT = '%Y-%m-%dT%H:%M:%S.%f+0000'
 DJANGO_DATETIME_FORMAT = '%Y-%m-%d %H:%M:%S.%f-00:00'
 
 request_count = 0
+table_map = {}
+
+
+def demap_table_names(query, q):
+	global table_map
+
+	for table, aliases in query.table_map.items():
+		mapping = table_map.setdefault(table, [])
+		mapping += aliases
+
+		for alias in mapping:
+			if alias is not table:
+				q = q.replace(alias, table)
+
+	return q
 
 def quoted_string_literal(s, d):
 	"""
@@ -90,7 +105,7 @@ def handle_api_exceptions(url, f, *args, **kwargs):
 		f:  requests.get or requests.post...
 		_cursor: sharing the debug information in cursor
 	"""
-	global request_count 
+	global request_count
 	from salesforce.backend import base
 	kwargs_in = {'timeout': getattr(settings, 'SALESFORCE_QUERY_TIMEOUT', 3)}
 	kwargs_in.update(kwargs)
@@ -396,6 +411,9 @@ class CursorWrapper(object):
 		Send a query to the Salesforce API.
 		"""
 		from salesforce.backend import base
+
+		if DJANGO_17_PLUS:
+			q = demap_table_names(self.query, q)
 
 		self.rowcount = None
 		if isinstance(self.query, SalesforceQuery) or self.query is None:


philchristensen:
So, just to be clear, this pull request is superseded by @hynekcer's #104, and can be closed, right?

Y3K:
@philchristensen @hynekcer thank you for your support! I made this more like a patch than an actual solution (since I don't know your codebase very well yet).

If you upgrade your API to support queries like

```
SELECT OpportunityLineItem.Id, OpportunityLineItem.OpportunityId FROM OpportunityLineItem WHERE (OpportunityLineItem.HasSchedule = %s AND OpportunityLineItem.OpportunityId IN (SELECT U0.Id FROM Opportunity WHERE (U0.StageName = %s AND U0.Probability > %s AND (U0.CloseDate >= %s OR U0.Name LIKE %s) AND U0.Amount > %s)))
```

I'd be more than happy either if the PR is merged or closed.

Thanks.

hynekcer:
The most important problem is that your patch doesn't work for me. (I tried it on OpportunityContactRole instead of OpportunityLineItem because the latter has so many read-only fields that the objects can not be created directly in automatic tests.) I see that table names are mapped to the identical names and nothind is demapped finally (The alias `U0` remains):

    >>> self.query.table_map
    {'OpportunityContactRole': ['OpportunityContactRole'], 'Opportunity': ['Opportunity']}

Do you need to add the mapping to the global variable `table_map` at the beginning manually?

Other problem is that seven tests start to fail with your patch, but it can be fixed by

    - if DJANGO_17_PLUS:
    + if DJANGO_17_PLUS and hasattr(self.query, 'table_map'):

Replacing all substrings like "U0" in the complete compiled SQL command is prone to bugs.

The last small problem is that it is not compatible with the fix for #43 in the PR 104.

----

More general also about my compiler fixes: An invalid query resulting with SalesforceError is currently still much more probable than an incorrect valid query with bad results, if the query is outside the implemented scope. This was good. I think that the NotImplementedError should be raised for too complicated cases that could be simpiified by any patch to anything valid but incorrect. It is frequently easier to write a correct complicated query manually than to manually verify the correctness of an automatically created query.

I think that you can compile an equivalent query with the latest master by

    OpportunityLineItem.objects.filter(has_schedule=..., opportunity__stage_name=...,
            opportunity__probability__gt=...,
            (Q(opportunity__close_categte=...) | Q(opportunity__mame__contains=...)),
            opportunity__amount__gt=...)

I want also to commit a few older changesets soon that allows to compile queries with a complicated relationship tree, e.g. a filter for a person by the FirstName of paternal grandmother and Birthdate of maternal grandfather: `filter(father__mother__first_name=..., mother__father__birthdate__gt=...)`

This issue can stay open, but it has a quite good workaround now. If it will be fixed ever, I think it will be done the simplest way only by a special new queryset method incompatible with other backends.

hynekcer:
Fixed. Very similar test is used. It should work also for you.

PR #104, opened 2015-05-20
Cumulative change: Queries based on more ForeignKeys to the same table etc

hynekcer:
This is a cumulative pull request with finished more than a half of my development since November.

- It is possible to write filters with more foreign keys related to the same model..
- More compatibility with standard Django + migrations (no real migrations, but very useful if the development is done on any standard Django database), serializable default values etc.
- Improved introspection, supported DEFAULTED_ON_CREATE, better comments to exported models, complete compatibility betwen Python 2 and 3.
- Fixed SSL protocol settings to current security requirements

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -110,7 +110,12 @@ variant until `django-salesforce>=0.5`.
 Foreign Key Support
 -------------------
 
-**Foreign key** filters are currently possible only for the first level of
+**Foreign key filters** are currently possible only from child to parent with some
+restrictions:
+
+They are fully supported with Django 1.7+ without other restrictions. **New**
+
+With Django 1.6 and older, ForeignKey filters are  possible only for the first level of
 relationship and only for fields whose name equals the name of object.
 Foreign keys of an object can be normally accessed by dot notation without any
 restriction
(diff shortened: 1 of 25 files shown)


hynekcer:
Everything from [Draft release notes](https://github.com/django-salesforce/django-salesforce/releases/tag/untagged-260e82c4b8aae5cf3d95) is also committed here. (I removed today obsoleted links to commit numbers because they have been reorganized since that.

philchristensen:
This all looks great. It passed all my tests okay as well.

This seems like a pretty large amount of important updates (especially the SSL ones), worth making a 0.6 release. Do you agree, @hynekcer ?

hynekcer:
A new version can be released soon e.g. next friday, but still without changing the whitespace everywhere. I still have aproximately fourty useful unmerged commits in seven branches. I should review them, combine the best together in a right order, to write tests for them and to begin to be sure it's so the right way. It will take about to the middle of July until I find slowly enough leisture time.

philchristensen:
that makes sense. glad to wait on the white space change, as it obviously affects only a tiny number of users.

PR #105, opened 2015-06-10
match converted_status to default schema

hoprocker:
I've been working against a default Salesforce install. It seems that the default picklist values for the Lead->LeadStatus field only include 'Contacted', 'Open', 'Qualified', 'Unqualified'. The error thrown was a bit cryptic:

```Lead conversion failed: Statusinvalid convertedStatus: Qualified - convertedINVALID_STATUS, leadId=...```

I eventually tracked it down as being a default value for the convert_lead() method. I wonder if this should be surfaced more visibly as a configuration parameter?

Btw, thanks for this library!

### DIFF
--- a/salesforce/utils.py
+++ b/salesforce/utils.py
@@ -19,7 +19,7 @@ except ImportError:
     beatbox = None
 
 
-def convert_lead(lead, converted_status="Qualified - converted"):
+def convert_lead(lead, converted_status="Qualified"):
     """
     Convert `lead` using the `convertLead()` endpoint exposed
     by the SOAP API.


philchristensen:
I think @tvanesse might be able to shed some light on whether this is intended behavior.

In the meantime, you may want to read over the discussion around this function's creation in #96

tvanesse:
@hoprocker Unfortunately the valid lead status are indeed changing over different Salesforce versions. So "Qualified - converted" used to work for me but it is very possible that another version throws the exception you just mentioned.

> I wonder if this should be surfaced more visibly as a configuration parameter

That is an idea but even better would be to use [`LeadStatus`](https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_objects_leadstatus.htm) in order to retrieve the available lead status at runtime and pick the one containing "qualified" or "converted".

Would you be willing to provide a PR that implements that?

tvanesse:
Still, I think it is a good idea to merge this PR already since the problem is likely to affect everyone who is using `convert_lead` today. But to avoid to have to manually adapt the code every time Salesforce decides to change the lead status, one could implement the mechanism described above.

hoprocker:
@tvanesse, I see the logic to your approach, but it seems to me there's still room for confusion; any organization that uses a custom picklist can have a case which breaks this.

This is a really cool feature, but my sense is that it shouldn't make assumptions about the picklist in question. One possibility that comes to mind is:

* converted_status can be provided as part of the function signature;
* if it's not provided, it looks for it in Django's settings (i.e. something like SALESFORCE_CONVERTED_STATUS);
* doesn't exist? error!!

What do you think?

philchristensen:
So far I'd just as soon merge this PR as is to reflect what new SF accounts get. Improvements are welcome, but I'd prefer something that uses introspection as opposed to configuration, if possible.

hoprocker:
(Just following up on this, sorry about the delay)

If we look at Opportunity.stage_name.choices and use the first one that matches against `/^Qualified.*/` -- or, if there's no match, use this first one available -- would this be complete?

hynekcer:
Fixed with introspection. Lead conversion also supports many optional parameters now.

Issue #108, opened 2015-06-21
Update README for PyPI support


philchristensen:
I had originally converted the README to Restructured Text because that's what PyPI supported, but it looks like they stopped, or something else is wrong:

https://pypi.python.org/pypi/django-salesforce

My best guess is they switched to using Markdown, but it's also possible we introduced a formatting bug into the README at some point. Or maybe need to use a different extension. Who knows...

hynekcer:
Comments in [this SF question](http://stackoverflow.com/q/28820818/448474) maybe can help.

Check the syntax by: `rst2html README.rst >something.html` or use

Distutils command to view and validate restructured text in package's long_description:
https://pypi.python.org/pypi/collective.checkdocs/

I wrote docs like a simple text with a decoration similar to the existing, without any knowledge about RST rules. I followed only correct color markup in my text editor. It seems that the only reported error is in a contribution of a third person.

philchristensen:
Closed by #113

PR #112, opened 2015-07-22
Fixed Readme syntax  #108

hynekcer:
Fixed Restructured Text syntax


hynekcer:
based on incorrect parent

PR #118, opened 2015-07-26
Configurable TLS. closes #109

### LINKED ISSUE #109: Salesforce Disabling TLS 1.0 Encryption (..in 2016)
Just cleaning out my email inbox and wanted to make sure this is on the radar.

[This email from Salesforce](https://gist.github.com/rwhitman/4728ef2788c860d4282c) June 15 alerts that they are phasing out TLS 1.0 support

Not super familiar with the internal workings of django-salesforce or python's ssl packages but took a spin and spotted `ssl_version = ssl.PROTOCOL_TLSv1` in [adapter.py](https://github.com/django-salesforce/django-salesforce/blob/master/salesforce/backend/adapter.py). Not sure if it defaults to 1.1 or it needs to be named explicitly? Would this be an issue?

hynekcer:
info in README

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -290,6 +290,42 @@ can see in the output of ``inspectdb`` in the most complete form.
    `instructions <http://www.python.org/https://help.salesforce.com/apex/HTViewHelpDoc?id=customize_mapleads.htm>`__
    for more details.
 
+SSL/TLS settings
+----------------
+The package `requests <http://python-requests.org>`__ doesn't allow easily to set
+the minimal required SSL/TLS version together with usage of the highest
+version that is available on both sides.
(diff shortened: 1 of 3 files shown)


philchristensen:
This looks great. I made some edits to the README text in ecd19257 let me know if there's any problems.

philchristensen:
I saw your draft release notes, but they imply the current default version is PROTOCOL_SSLv23, but from the code it looks to me like it defaults to PROTOCOL_TLSv1 (I added something to that effect to the README as well).

hynekcer:
The highest version can be enabled by settings `METHOD_SSLv23`. I explaind the setting instead of qoute it

PR #119, opened 2015-08-04
Some fixes for Django 1.8, Tests for SSL, filter with similar FK etc

hynekcer:
Much documentation together with code

If the test of TLS 1.1+ fails and no upgrade is possible, it can be skipped by
settings.py

       SF_SSL = {'skiptest_tls_11': True, ...}

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -322,6 +322,17 @@ the old insecure protocols including SSL v3 are disabled unless you've installed
 PyOpenSSL. As long as you have *not* installed PyOpenSSL, it's recommended you
 update your settings to use `PROTOCOL_SSLv23`.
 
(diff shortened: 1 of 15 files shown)


philchristensen:
I'm seeing two relatively minor test failures:
t

philchristensen:
I see one of them is a config issue in my Travis setup...

hynekcer:
The second can be fixed by settings.py

    SF_SSL = {'ssl_version': ssl.PROTOCOL_SSLv23}

philchristensen:
Yep, just added that. It looks like the first is trivial as well. I'll make a PR for the hotfix.

PR #120, opened 2015-08-05
Travis config fix, plus a minor inspectdb fix

philchristensen:
The inspectdb fix was just a small change to the pop() usage


philchristensen:
I see what you mean about the ssl line. It shouldn't be a problem to move it to a local settings file.

PR #122, opened 2015-08-05
Run tox on the Travis build, to test multiple Python/Django versions

philchristensen:
We've been running tox internally for some time, but I don't think there's any reason I couldn't get it running on Travis. We can't use the built-in multi-environment support from Travis because it tries to run them all in parallel, which isn't currently supported by our tests if they're using the same account.

This pull request isn't ready for merge yet, but I have a feeling I may have some questions for @hynekcer and this will be a good place for discussion.

Incidentally, the reason this is on a branch on the main repo instead of on my personal fork is because the encrypted environment variables in the .travis.yml file are essentially keyed to this repo, so only pushes from the repo will be able to access the demo Salesforce account I use to test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,5 @@
 sudo: false
 language: python
-python:
-  # - "2.6"
-  - "2.7"
-  # - "3.3"
 env:
   global:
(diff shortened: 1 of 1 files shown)


hynekcer:
If the exact version of Python is also a problem, I can remove it from tox.ini and leave only "python2.7" etc. again. The latest versions can be expected on Travis. It hasn't been important before writing SSL tests.

PR #123, opened 2015-08-06
add none() method to SalesforceManager

chromakey:
Prior to this patch, if you ran the none method on a salesforce model, it returned a salesforce error.  

i.e. Lead.objects.none() returned a salesforce error.

The salesforce error returned is:

salesforce.backend.base.SalesforceError
salesforce.backend.base.SalesforceError: SOQL statements cannot be empty or null

This permits you to use the none() method on salesforce models.

This is my first pull request to a public repo ever, so apologies in advance.

### DIFF
--- a/salesforce/backend/manager.py
+++ b/salesforce/backend/manager.py
@@ -64,3 +64,6 @@ class SalesforceManager(manager.Manager):
 			return self.get_queryset().query_all()
 		else:
 			return self.get_queryset()
+
+	def none(self):
+		return super(SalesforceManager, self).get_queryset().none()


philchristensen:
This looks good to me, should be able to merge soon.

hynekcer:
It must work also inside the queryset. The module salesforce.backend.manager is not the right place where to fix this bug. If you try e.g.

    Lead.objects.all().none()

you get the same error as before the fix.

What is your Django version and django-salesforce? I can not reproduce your error on the current public master branch. Django 1.4 and 1.5 works correctly. Django 1.6 - 1.8 raises other error EmptyResultSet

chromakey:
I'm running Django 1.7.8, python 3.4, and django-salesforce 0.6.

You are indeed correct that
```
Lead.objects.all().none()
```
returns the same error.  I will investigate further.

hynekcer:
Do you get SalesforceError? This is the same EmptyResultSet on my computer. Can you reproduce it on the test project? Unpack the zip from pypi. Copy your database settings to salesforce/testrunner/example/local_settings.py  and run `python manage.py shell` the the base directory of the package.

what is the result of

    Lead.objects.none().query.get_compiler('salesforce').as_sql()

chromakey:
I'm going to close the pull request and open an issue.  If we figure out that a patch is indeed needed, I can open a new pull request.

Issue #124, opened 2015-08-07
none() method doesn't work on Salesforce models


chromakey:
**This issue continues from PR #123** (edit: hynekcer)

Using the instructions from @hynekcer , I get a raise EmptyResultSet running both of the queries below on the test project:

```
Lead.objects.none()
Lead.objects.none().query.get_complier('salesforce').as_sql()
```

When I add my patch to the zip from pypi:

```
Lead.objects.none()
returns
[]
```

```
Lead.objects.none().query.get_compiler('salesforce').as_sql()
returns
raise EmptyResultSet
```

The project I'm working on where I stumbled upon this is a very basic CRUD front end for Salesforce leads, nothing interesting at all.  When I install my patch, it exhibits the same behavior as when I apply it to the test project.  

I'm using the Lead.objects.none() query in conjunction with a **django-filter**.  It works fine with non-salesforce models.

```
    filter = UserForm(queryset=User.objects.none())
    paginator = Paginator(filter.qs, 10)
    page = request.GET.get('page')

    try:
        paged_leads = paginator.page(page)
    except PageNotAnInteger:
        paged_leads = paginator.page(1)
    except EmptyPage:
        paged_leads = paginator.page(paginator.num_pages)
```

The above works fine using the django.contrib.auth User model.  However, the below returns a "salesforce.backend.base.SalesforceError: SOQL statements cannot be empty or null" error:

```
    filter = LeadFilter(queryset=Lead.objects.none())
    paginator = Paginator(filter.qs, 10)
    page = request.GET.get('page')

    try:
        paged_leads = paginator.page(page)
    except PageNotAnInteger:
        paged_leads = paginator.page(1)
    except EmptyPage:
        paged_leads = paginator.page(paginator.num_pages)

```
The forms are just simple
```
class UserForm(django_filters.FilterSet):
    class Meta:
        model = User

class LeadFilter(django_filters.FilterSet):
    class Meta:
        model = Lead
```
I am running django-salesforce 0.6, Python 3.4, and Django 1.7.8.

hynekcer:
Please look at your traceback, find the line with the problematic queryset and put a debug print before it:

    print(my_queryset.query.get_compiler('salesforce').as_sql())

Repeat it until you find two lines that can be easily reproduced from django shell command line.
Your error is raiser between lines 39 - 49 in backend/query.py  
You can call debugger just before line 139

    newline   import pdb; pdb.set_trace()
[...]

PR #126, opened 2015-08-08
Many2many relationship, advanced Where conditions

hynekcer:
- Implemented Many2many realtionship so that the related name .all() and the reverse related name .all() works, with Django 1.4 to 1.8

- Speed up tests. The connection to SFDC is not created until necessary, because some tests don't require the connection at all.

closes #103, #124

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -171,8 +171,8 @@ Advanced usage
    
    One way to speed this up is to change the SALESFORCE_DB_ALIAS to point to
    another DB connection (preferably SQLite) during testing using the
-   ``TEST_*`` settings variables. The only outbound connections will then be to
-   the authentication servers.
+   ``TEST_*`` settings variables. Django unit tests without SalesforceModel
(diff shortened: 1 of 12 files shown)


philchristensen:
Wow! Great stuff.

philchristensen:
I'm having some issues with the tests on Travis, but they all appear related to the fact that I ended up running two builds at once because of two merges right in a row. I'll fix the tests ASAP.

hynekcer:
The object `django_Test__c` is the model `example.models.Test`. It is really a bug that the test is not skipped. I recommend to create soon if you want to test relationships between custom and builtin objects (in the future, currently it is skipped because wuold fail) It is currently the only custom object that is really tested on SFDC. (custom object `ChargentOrders__ChargentOrder__c` is "mocked")

philchristensen:
Thanks for your help. Tests are passing fine on the Travis build now.

PR #127, opened 2015-08-08
Fixed query compiler

hynekcer:
I think it is an important fix and a big part of issues is now ready. The standard result is currently 5 skipped tests for SSL.

I also have sometimes the problem that while long queue of tests is running I need to try one short test. I improved it again.

Other important fix will be query_results method of query. I can continue in September.


philchristensen:
Are we good to do a 0.6.1 release now (after a few days to let things settle?)

hynekcer:
Maybe, but I will continue in September

PR #131, opened 2015-09-04
add range lookup functionality

chromakey:
This pull request modifies the range lookup provided by Django to render the correct SOQL code for the Salesforce API.  The reason this is needed is that Salesforce does not recognize the BETWEEN clause in SOQL.

Prior to this pull request:
```
today = datetime.date.today()
yesterday = today - datetime.timedelta(days=1)
opportunities = Opportunity.objects.filter(close_date__range(yesterday, today))
```
generated this SQL:
```
'SELECT <<my_opportunity_fields>> FROM Opportunity WHERE Opportunity.close_date BETWEEN %s AND %s', (datetime.date(2015, 9, 1), datetime.date(2015, 9, 2))
```
This pull request now renders this SOQL:
```
'SELECT <<my_opportunity_fields>> FROM Opportunity WHERE Opportunity.close_date >= %s AND Opportunity.close_date <= %s', (datetime.date(2015, 9, 1), datetime.date(2015, 9, 2))
```
I used the work done on the isnull lookup and this link as references.  

https://docs.djangoproject.com/en/1.8/howto/custom-lookups/#writing-alternative-implementations-for-existing-lookups

I wrote a test for it as well.  Let me know what you think.  Thanks!

### DIFF
--- a/salesforce/backend/compiler.py
+++ b/salesforce/backend/compiler.py
@@ -601,8 +601,8 @@ if DJANGO_17_PLUS:
 
 	models.Field.register_lookup(IsNull)
 
-
-	class Range(models.Field.get_lookup(models.Field(), 'range')):
+        BaseRange = models.Field.get_lookup(models.Field(), 'range')
+	class Range(BaseRange):
 		# The expected result base class above is `models.lookups.Range`.
 		lookup_name = 'range'
 
--- a/salesforce/backend/compiler.py
+++ b/salesforce/backend/compiler.py
@@ -9,6 +9,7 @@
 Generate queries using the SOQL dialect.
 """
 import re
+import datetime
 from django.db import models
 from django.db.models.sql import compiler, query, where, constants, AND, OR
 from django.db.models.sql.datastructures import EmptyResultSet
@@ -609,7 +610,14 @@ if DJANGO_17_PLUS:
 			if connection.vendor == 'salesforce':
 				lhs, lhs_params = qn.compile(self.lhs)
 				rhs, rhs_params = self.process_rhs(qn, connection)
-				return '%s >= %s AND %s <= %s' % (lhs, rhs_params[0], lhs, rhs_params[1]), lhs_params
+
+				if type(rhs_params[0]) is str and type(rhs_params[1]) is str:
+					return "%s >= '%s' AND %s <= '%s'" % (lhs, rhs_params[0], lhs, rhs_params[1]), lhs_params
+				elif type(rhs_params[0]) is datetime.datetime and type(rhs_params[1]) is datetime.datetime:
+					return "%s >= %s AND %s <= %s" % (lhs, rhs_params[0].strftime("%Y-%m-%dT%H:%M:%S.%fZ"), lhs,
+													  rhs_params[1].strftime("%Y-%m-%dT%H:%M:%S.%fZ")), lhs_params
+				else:
+					return "%s >= %s AND %s <= %s" % (lhs, rhs_params[0], lhs, rhs_params[1]), lhs_params
 			else:
 				return super(Range, self).as_sql(qn, connection)
 
--- a/salesforce/backend/compiler.py
+++ b/salesforce/backend/compiler.py
@@ -381,10 +381,6 @@ class SalesforceWhereNode(where.WhereNode):
 				return ('%s %snull' % (field_sql,
 					(not value_annot and '!= ' or '= ')), ())
 
-			if lookup_type == 'range':
-				return ('%s >= %s AND %s <= %s' % (field_sql,
-												   value_annot[0], field_sql,
-												   value_annot[1]), ())
 		else:
 			return result
 
--- a/salesforce/backend/compiler.py
+++ b/salesforce/backend/compiler.py
@@ -345,7 +345,7 @@ class SQLCompiler(compiler.SQLCompiler):
 
 
 class SalesforceWhereNode(where.WhereNode):
-	overridden_types = ['isnull', 'range']
+	overridden_types = ['isnull']
 
 	# Simple related fields work only without this, but for more complicated
 	# cases this must be fixed and re-enabled.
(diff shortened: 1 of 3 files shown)


philchristensen:
Looks good so far. My main question is how does it handle range requests for non-date data types, and does its behavior/error handling for other types make sense?

It's fine if it only supports a limited subset of data types, but I'd like to know it would give an intelligent error for unexpected objects.

chromakey:
Range now supports Date, DateTime, Decimal/Integer (should do Float as well), and CharField.  Tests included.

philchristensen:
That's great. If it behaves essentially the same as Django I'm happy.

salesforce/backend/compiler.py
```diff
@@ -596,6 +601,28 @@ def as_sql(self, qn, connection):
 
 	models.Field.register_lookup(IsNull)
 
+
+	class Range(models.Field.get_lookup(models.Field(), 'range')):
```
philchristensen: Not to be pedantic, but I feel like this syntax is a little un-Pythonic. What do you think about doing something like:

```python
BaseRange = models.Field.get_lookup(models.Field(), 'range')
class Range(BaseRange):
    ...
```

philchristensen:
Thanks. I'm okay with the merge if @hynekcer is.

hynekcer:
I appreciate very much that you added the link to Custom Lookups docs to your code. That explains that the recommended way is simple. I still don't understand why Django uses the syntax "BETWEEN ... AND ..." twice: once in "where.WhereNode.make_atom" and once in "lookups.Range". I am afraid that maybe only some cases are fixed or they may had a bug in Django and they fixed it by a duplicated code?

Is it useful to add the string `'range'` to `overridden_types` in your code?

The third commit will fail in Python3 due to mixed spaces and tabs, but I prefer now a simple static base similarly to the docs. I will modify the test to be faster and to test what philchristensen has meant. Also expectedFailure should be marked if Django is older than 1.7.

.

chromakey:
OK I think I took care of all the requested changes.  I ran the test in python3 and it seemed to work fine after I fixed a datetime issue.  I think I scrubbed out the space/tab mixing.  

I'm not sure how I should trigger an alert to a non-Django 1.7+ user if they try to run the range lookup.  The range lookup is only overridden in the code if the user is on Django 1.7+ or it reverts to pre-existing behavior.  If you want to point me in the right direction I can take care of it.  1.4 is EOL'd on October 1, so not sure if that factors into the decision or not.

philchristensen:
I'm all about dropping support for unsupported django-salesforce versions.

philchristensen:
Er, I meant Django versions

chromakey:
Out of curiosity, why was this closed without merge?  Was something missing?

philchristensen:
whoa, really? must have been a mistake.

philchristensen:
Oh shit, yeah, that was me. My bad. Thanks again for the contribution!

hynekcer:
I'm all for it. Django 1.4 LTS is ending on October 1st.Then we remove much code. Now we can remove the configuration "py26dj14" from automatic tox tests: `envlist = py27dj17, py33dj17, py27dj18` and this can be merged immediately without waiting until I have time.

chromakey:
Thanks guys! Thanks for all of your work on this great library.

philchristensen:
Thanks for your contributions.This is my first really successful Open Source project, and a huge amount of work and fixes and features have come from @hynekcer who has been making killer contributions for years. I'm finally starting to use this library again at my latest job, and I'm so pleased to have such a nice community to help test and implement this stuff.

philchristensen:
This is looking pretty good, just one question about a unit test:
```python
======================================================================
ERROR: test_range_lookup (salesforce.tests.test_integration.BasicLeadSOQLTest)
Get the test opportunity record by range condition.
----------------------------------------------------------------------
  File "/home/travis/build/django-salesforce/django-salesforce/salesforce/tests/test_integration.py", line 778, in test_range_lookup
    oppy = Opportunity.objects.filter(created_date__range=(start_time, end_time))[0]
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/db/models/query.py", line 177, in __getitem__
    return list(qs)[0]
IndexError: list index out of range
----------------------------------------------------------------------
```

Do I need to create a record in my test org or something to make this work?

chromakey:
Hmmm.  I create an opportunity in the setUp portion of the test and everything passed through accordingly for me, so I'm not sure.

philchristensen:
No prob. It could be something in my travis setup, it's still a little janky right now.

hynekcer:
@chromakey Please fix the test. The cause of problem is `RuntimeWarning: DateTimeField Opportunity.created_date received a naive datetime (...) while time zone support is active`. The result is that the test fails in America and in the whole Western Hemisphere.

hynekcer:
I found a serious bug in your code that allows "SQL Injection" vulnerability.

     >>> Opportunity.objects.filter(name__range=("\\", 'z'))
    SalesforceError: MALFORMED_QUERY
    >= '\' AND Opportunity.Name <= 'z' LIMIT 21
                                   ^
    unexpected token: 'z'

    >>> Opportunity.objects.filter(name__range=("'", 'z'))
    SalesforceError: MALFORMED_QUERY
    Opportunity WHERE Opportunity.Name >= ''' AND Opportunity.Name <= 'z' LIMIT
                                           ^
    unexpected token: ' AND Opportunity.Name <= '

This can cause a data leak, though it is not easy in a very simple query thanks to strange SOQL syntax, because parentheses **must** be used for valid SOQL if two different logical operators AND and OR are combined in the WHERE expression. The bug can not modify the the database.

The merge should be immediately reverted until it is fixed.

To the bug in the test: It will be useful also for you to use date/datetime range correctly. You can have the same problem with other databases, not only in Salesforce.

hynekcer:
I suggest to release "v0.6.1" now. This will be last version that support Django 1.4, 1.5, 1.6

I am going to remove the old code. It was many times easy to keep the old code, but it is more complicated for new contributors and it is not nice to see that the old code is copy pasted unnecessarily.

hynekcer:
Fixed again by 5fa46c5fe and b81db401b.

chromakey:
Thanks @hynekcer .  I've been crushed this past week and was going to try to tackle the problem this weekend, but you already got to it!

PR #132, opened 2015-09-12
Revert "add range lookup functionality"

hynekcer:
Reverts django-salesforce/django-salesforce#131
(see the comment about SOQL injection)


philchristensen:
Thanks for catching this. I'm also going to start using an alternate branch to test PRs in the future, don't know why I didn't do that in the first place

PR #134, opened 2015-09-28
Documentation Cleanup (in progress)

philchristensen:
Our README has been our only formal source of documentation for this project for some time, but it's been growing quite substantially, particularly as @hynekcer has plumbed some of the deepest depths of the SFDC realms. This is fantastic knowledge to have collected in one place, but most of it is increasingly unnecessary to first-time users.

I've started the process of moving some of the more esoteric content out of the README, and into the Wiki. I'm not really sure when this will hit `master`, but I would hope soon.

Any input or suggestions would be appreciated,

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -108,58 +108,6 @@ primary key `id` by configuring `SF_PK='id'` in your project settings. The previ
 capitalization of `Id` is only for old projects, but it will stay as the default
 variant until `django-salesforce>=0.5`.
 
-Foreign Key Support
--------------------
-
-**Foreign key filters** are currently possible only from child to parent with some
-restrictions:
-
-They are fully supported with Django 1.7+ without other restrictions. **New**
-
-With Django 1.6 and older, ForeignKey filters are  possible only for the first level of
-relationship and only for fields whose name equals the name of object.
-Foreign keys of an object can be normally accessed by dot notation without any
-restriction
-Example::
-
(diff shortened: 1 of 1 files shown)


hynekcer:
I agree with you. It is nice that the wiki can be cloned and searched locally by `git clone https://github.com/django-salesforce/django-salesforce.wiki.git`.

I would like also to split example.models later to
- nomal examples that can be used in Readme examples (some also in simple tests)
- special and strange models for more complicated tests, e.g. backward compatibility tests

Some example models use lower_case field names, some use MixedCase and some use both. Lower case are fine with ORM, but problematic if also raw queries must be hand written. Custom fields are really a big problem how to write raw query correctly by hand and get the result from cursor.

A fix require to wait for some work and decisions:
- Split the query module to a DB API driver and Django backend query.
- Implement a normal cursor that returns primarily a list of values (what expects Django even for related fields), not only a tree of dictionaries as it is currently.
- Compiler of dot separated ORM names to SFDC names and a reverse compiler, including correct "__c" and "__r" suffixes. The reverse compiler should be case insensitive, otherwise raw queries are easily broken.
- Fix raw queries for Django 1.8

Without these points I can't say if MixedCase names should be definitively obsoleted or not yet.

Currently I have 20 new commits, written immediately after v0.6.1 (e.g. bulk insert, bulk update and detection of master detail relationship) but I have no time to refactor it nicely. These commits have been inspired by reading "Caveats" in Readme.

PR #137, opened 2015-12-08
Fixed for Django 1.9, master-detail relationship, etc

hynekcer:
- Added support for Django 1.9 in the same range as for Django 1.8 (little less than for Django 1.7)
- Removed old code for Django 1.4 - 1.6 and Python 2.6, fixed deprecation warnings.
- Inspectdb recognizes "master-detail relationships" from "lookup fields".
- Updated to Salesforce Winter '16 API v35.0.


hynekcer:
Fixed inspectdb for Django 1.9

What do you thing about the initial warning for Django 1.8 and 1.9?

hynekcer:
Yes, your text is good. I mean also the current problematic form that it is a RuntimeWarning that people see after "import salesforce". I know that I wrote it, but now I asked myself: Isn't anything broken with our package? Why it looks like a short traceback? May I still recommend a version 1.7 unsupported by Django developers? I am inclined better to remove the message.

philchristensen:
Oh, I understand. Yeah, it might be just as well to remove the warning; it's still highly functional in these other versions, well tested, and probably meets 90% of the use cases.

philchristensen:
This looks ready to merge to me, or would you rather remove that RuntimeWarning first?

hynekcer:
Please remove it and maybe also fix anything in README. I am busy now, fixing our broken server on other end of the republic and after my normal job expected next week I'm going to my city only next weekend. However I'm glad that we made the 1.9 very soon after the Django release.

philchristensen:
No problem, I got it. Thanks again for all your excellent work.

PR #140, opened 2015-12-26
Migrate some README to Wiki

hynekcer:
I found a bug that [our README on pypi](https://pypi.python.org/pypi/django-salesforce/0.6.2) is without formating again due to invalid syntax caused by some missing whitespace. I really should not edit the README so frequently as now.

I rebased my changes to README made in September by me concurrently with your #134. I have merged both now.

I also copied the shaphost of github.com/django-salesforce Wiki to `./wiki` directory. I need it for the development of new features in development branches until they are ready to be published and merged to master. Wiki pages for the current code can be naturally written and fixed directly by github web. I will synchronize them before any push from wiki directory to wiki web.


philchristensen:
Thanks, the README really was getting unwieldy.

Issue #71, opened 2014-08-07
App failure due to SFDC connectivity issues, even for non-SFDC functions


naoyak:
Yesterday I had some problems on my app due to service disruptions on the Salesforce server instance hosting our data, which caused authorization to fail and trigger the `LookupError` in `auth.authenticate`. Are there any recommended solutions for handling this exception in order not to disrupt app functionality (aside from calls to Salesforce)? In other words, when SFDC service is disrupted I'd like SF-related features to fail quietly without having an exception take down the app.

PR #146, opened 2016-02-10
Mention SF_LAZY_CONNECT settings in README.rst

MounirMesselmeni:
This settings is important to mention for users specially to avoid problems when the application don't need to communicate with Salesforce and salesforce is down or in maintenance.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -102,6 +102,11 @@ Quick Start
     # ...
     # end of file
     register_omitted_classes(your_application.models)
+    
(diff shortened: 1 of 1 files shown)


hynekcer:
I agree. This is exactly in the intentions of the original issue #71 . It is useful for a server while the package can start and to recover later from the disconnected state. This setting is probably not useful for development or even not for quick start with the package.

hynekcer:
@philchristensen Please write something clear like "but don't set that unit you verify that the login to Salesforce is correctly configured".

(The reason is that the user account can be more easily locked by repeated requests if the application continues, although the password or security token are invalid.)

Issue #56, opened 2014-05-29
application unittest strategy


keturn:
Our current strategy for automated unit tests in our django project relies on using an in-memory sqlite3 database, replacing the default engine with `django.db.backends.sqlite3`.  This isn't perfect, as it doesn't exactly match postgresql semantics, but it seems to be generally accepted practice.

How do we write unit tests for our application that uses a django-salesforce database?

I imagine we could start by having our test settings use sqlite3 for the salesforce backend as well, or by removing the salesforce ModelRouter from our test settings.

Points to address include:
1. The salesforce models are not managed. How to have the test runner create tables for them?
2. According to README's "Foreign Key Support" (and other caveats), there are queries the sqlite3 backend will support that the salesforce backend will not. How to avoid writing tests that will pass in sqlite3 but fail when run in production against salesforce?

I'll probably also want a test suite that actually runs against a real salesforce backend, but due to speed and API limits I don't think I'll want to rely on that approach exclusively.

PR #149, opened 2016-02-20
Data migrations broken since 0.6.2

lociii:
Data migrations don't work on a model base but the salesforce router always checks for model attributes.
Therefore RunSQL and RunPython are broken when using the salesforce router.

> Running migrations:
>   Rendering model states... DONE
>   Unapplying actionstream.0005_product_created...Traceback (most recent call last):
>   File "./manage.py", line 11, in <module>
>     execute_from_command_line(sys.argv)
>   File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
>     utility.execute()
>   File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 346, in execute
>     self.fetch_command(subcommand).run_from_argv(self.argv)
>   File "/usr/local/lib/python2.7/dist-packages/django/core/management/base.py", line 394, in run_from_argv
>     self.execute(*args, **cmd_options)
>   File "/usr/local/lib/python2.7/dist-packages/django/core/management/base.py", line 445, in execute
>     output = self.handle(*args, **options)
>   File "/usr/local/lib/python2.7/dist-packages/django/core/management/commands/migrate.py", line 222, in handle
>     executor.migrate(targets, plan, fake=fake, fake_initial=fake_initial)
>   File "/usr/local/lib/python2.7/dist-packages/django/db/migrations/executor.py", line 112, in migrate
>     self.unapply_migration(states[migration], migration, fake=fake)
>   File "/usr/local/lib/python2.7/dist-packages/django/db/migrations/executor.py", line 168, in unapply_migration
>     state = migration.unapply(state, schema_editor)
>   File "/usr/local/lib/python2.7/dist-packages/django/db/migrations/migration.py", line 162, in unapply
>     operation.database_backwards(self.app_label, schema_editor, from_state, to_state)
>   File "/usr/local/lib/python2.7/dist-packages/django/db/migrations/operations/special.py", line 188, in database_backwards
>     if router.allow_migrate(schema_editor.connection.alias, app_label, **self.hints):
>   File "/usr/local/lib/python2.7/dist-packages/django/db/utils.py", line 361, in allow_migrate
>     allow = method(db, app_label, **hints)
>   File "/usr/local/lib/python2.7/dist-packages/salesforce/router.py", line 74, in allow_migrate
>     model = hints['model'] if 'model' in hints else apps.get_model(app_label, model_name)
>   File "/usr/local/lib/python2.7/dist-packages/django/apps/registry.py", line 201, in get_model
>     app_label, model_name = app_label.split('.')
> ValueError: need more than 1 value to unpack
> 

Please make a new release as soon as possible. Current one is broken and prevents users from working with data migrations.

### DIFF
--- a/salesforce/router.py
+++ b/salesforce/router.py
@@ -71,6 +71,11 @@ class ModelRouter(object):
         """
         Don't attempt to sync SF models to non SF databases and vice versa.
         """
+
+        # data migrations can be skipped
+        if 'model' not in hints and not model_name:
+            return
+
         model = hints['model'] if 'model' in hints else apps.get_model(app_label, model_name)
         if hasattr(model, '_salesforce_object'):
             # If SALESFORCE_DB_ALIAS is e.g. a sqlite3 database, than it can migrate SF models


philchristensen:
I don't see much issue with this, but what are RunSQL or RunPython?

Also, why would you even have data migrations in an app using django-salesforce? Or is this a scenario where the app has both standard and Salesforce models and you've applied data migrations to the non-SF models?

We can probably get the patch merged this weekend, but I don't have a specific timeline for the next release. But patience is a virtue.

hynekcer:
Can you please write more details? Especially: 
- Django version. 
- SALESFORCE_DB_ALIAS (default or salesforce?) 
- Do you have any migration file for models with base class SalesforceModel? 
- Can you reproduce that the problem is not with django-salesforce v0.6.1? 
- Did you migrated between Django versions in the history of the failing project?  
- I don't see that your patch skips data migrations. In the contrary, it only skips the result "don't migrate" (return False) in some cases. Your return value None means that the router has no opinion (not yes or no).

I can not reproduce your bug. Migrations seems that they work in the very limited extent that is implemented, that is described in README and that has been asked by users in any previous issue, if I remember. I will explain it more verbosely.

The attitude to migrations in Salesforce was initially: "Do nothing really, mainly don't cause problems." The issue #56 explained us that people need also fast tests without SFDC only with sqlite3 in the memory. This was the main purpose of migrations. If you remember Django 1.7, migrations were in about ten times slower there than in Django 1.8 or newer. Upgrading any project with Salesforce migrations from Django 1.7 to Django 1.8 was also problem under some circumstances. I have not heard yet that someone really uses other database for SalesforceModel, except for fast tests. We decided later not to allow to create migration files. It was OK until Django 1.9 where no table is created in alternative databases for SalesforceModel, though everything works with sqlite3 in tests. As I explained, I'm not sure that it is really desirable, maybe after cutting support for Django 1.7 or so. I expected that other databases would work permanently as a not yet required bonus, fortunately I have never claimed it explicitly in the documentation.

lociii:
Hi all. Thanks for your fast response.
Here are the requested details.

Django version: 1.8.9
SALESFORCE_DB_ALIAS: not set
Do you have any migration file for models with base class SalesforceModel: only initial migration, but tables aren't managed (managed=False)
Can you reproduce that the problem is not with django-salesforce v0.6.1: yes
Did you migrated between Django versions in the history of the failing project: yes

```python
DATABASE_ROUTERS = [
    'salesforce.router.ModelRouter',
]
```

The failing migration has nothing to do with salesforce models.
Django calls all database routers to check if a migration can be run. For data migrations, it will not send a model_name to the allow_migrate function of the router. In [router.py:74](https://github.com/django-salesforce/django-salesforce/blob/master/salesforce/router.py#L74) you're calling apps.get_model which expects either a dot-separated app/model string or app_name and model_name. This function call issues the error.

For information about RunSQL and RunPython, please see https://docs.djangoproject.com/en/1.9/topics/migrations/#data-migrations

hynekcer:
Ahh, there was an untreated empty input. Clearly. Thank you much for your patch.

I'm improving it a little for the case `managed = True`, because Salesforce should not run neither data migrations automatically (especially until the Bulk API will be merged), other databases can do it. Imagine migration of a table with 50000 rows by one row, then you exhaust the governor limits and would be locked for 24 hours.

lociii:
Thanks, guys!

philchristensen:
@hynekcer do you have any thoughts on when we should do another release? From what I can see, this is the first actual code change since 0.6.2.1, but there's several doc changes and the switch to soft tabs, which could probably justify at least a 0.6.2.2 release.

philchristensen:
Well, I was just bit by this bug in my own project. I'm going to go ahead with a 0.6.2.2 release.

PR #157, opened 2016-05-26
Static version string 

### LINKED ISSUE #156: Detection of package version
Four very different version numbers can be read. example: 

    git describe --tags        # 0.6.2.3-1-g78cfc73  (ok)
    python setup.py --version  # 0.5-30-g3a47a5b  (because I forgot to delete the file VERSION)
    pip show saleforce         # or pip freeze (invalid because not used "python setup.py develop")
    git describe               # v0.1.3-615-g78cfc73  (the last annotaded tag 615 commits ago)

1) The cached file `VERSION` is problematic becase it is not under version control and must be deleted regularly. It is so precise number that it is always unvalid under development process.

2) The second number is important because it can be reported incorrectly by the new `__version__` attribute unrelated to the source code.

3) The third problem need not be solved now. Maybe it can be fixed in the next releases by using `git tag -a  the_tag`. The option `-a` is intended for annotated tags, i.e. public releases, while normal tags are for private purposes. If you create tag releases by Github now, it should be good enough.

Discussion:
If I remember, such complicated scripts like "get_tagged_version" where the version is derived from a tag were important for Mercurial, due to chicken - egg problem, because [it is impossible in Mercurial to simultaneously commit and add a tag for the changeset being committed](https://www.mercurial-scm.org/wiki/Tag#How_do_tags_work_in_Mercurial.3F). Therefore the only exact solution was possible without writing the tag into the source code. This is easy with Git (and many other VCS) and I don't see a reason more. You can use a script (or a paper list) at your side to not forget anything, but a static version number should be present without dependency on *.dist-info, *.egg-info and similar info. If the user installs django-salesforce by `-e git+https://...` option,  she can not reliably report the installed version after running git pull.

Possible solutions:

Static version number in the code.

1a) The dependency between source code and git can be **[reversed](http://stackoverflow.com/questions/6786555/automatic-version-number-both-in-setup-py-setuptools-and-source-code)**.

    git tag -a v$(python setup.py --version)

1b) Releases can be created manually by Github.
It is useful that a pre-release version can be simply marked by writing a new `__version__ = ...-dev` to the code, without adding a tag too soon to invalid code. Finally to remove "-dev" and release a tag. Also unofficial micro-numbers can be useful.

2) The version number for `setup.py` can be read by parsing `__init__.py` if the git repository is not present. The current long version number is very good for an installation from git.

(I will continue.)

hynekcer:
Fixed #156. Do you agree with a static version number independent on setuptools, easy_intall, pip, setup.py etc. issues?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-django>=1.7
+django>=1.7,<1.9.99
 simplejson>=2.5.0
 pytz>=2012c
 requests>=2.0.0
--- a/salesforce/__init__.py
+++ b/salesforce/__init__.py
@@ -10,13 +10,13 @@ A database backend for the Django ORM.
 
 Allows access to all Salesforce objects accessible via the SOQL API.
 """
-from pkg_resources import get_distribution
 import logging
 import warnings
 
 import django
 
-__version__ = get_distribution('django-salesforce').version
+__version__ = "0.6.3"
+
 DJANGO_18_PLUS = django.VERSION[:2] >= (1, 8)
 DJANGO_184_PLUS = django.VERSION[:3] >= (1, 8, 4)
 DJANGO_19_PLUS = django.VERSION[:3] >= (1, 9)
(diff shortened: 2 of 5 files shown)


philchristensen:
I'm fine with it, since the Git tags seem more trouble than they're worth.

PR #158, opened 2016-06-27
Fixing SSL-TLS article link

markgarg:
The old URL pointed to a wrong article. It opens up a new page. Fixed the link.


philchristensen:
Thanks very much!

hynekcer:
Thanks for your previous fix. I am sorry, I am removing this link and the related code in the new versions 0.6.9 and 0.7, because all development sites must use TLSv1.1 or higher, even that production sites can still disable a critical update and use an old TLS. That means every developer must use a Python compiled with libssl 1.0.1 or newer and no important security hole can be patched more by the old way.

Issue #175, opened 2017-03-01
Issues and limitations around `select_related` usage


cooncesean:
## Context
We're attempting to reduce our overall network calls to SFDC. We are trying to do this by using `select_related()` to prefetch related objects associated to the model in our queryset.

## Example
For example, lets attempt to iterate over all `Opportunities` in our SFDC instance. For each `Opportunity`, print the associated opp owner (which is an `User` object). Assume the model definition below:

**models.py:**

```python
from salesforce.models import SalesforceModel

class SFDCUser(SalesforceModel):
    """Access to the SFDC `User` table."""
    username = models.CharField(max_length=80)

    class Meta(SalesforceModel.Meta):
        db_table = 'User'

class Opportunity(SalesforceModel):
    """Access to the SFDC `Opportunity` table."""
    owner = models.ForeignKey('SFDCUser', custom=False, db_column=u'OwnerId')
  
    class Meta(models.Model.Meta):
        db_table = 'Opportunity'
```

**query.py:**

```python
# Fetch each opp while joining on owner to avoid subsequent calls to 
# get owner by id for each iteration.
for opp in Opportunity.objects.all().select_related('owner'):
    print opp.owner

---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-91-bdc76891f5b9> in <module>()
----> 1 for o in opp:
      2     print o
      3     print o.owner
      4

/usr/local/etc/virtualenvs/gears/local/lib/python2.7/site-packages/django/db/models/query.pyc in __iter__(self)
    256                - Responsible for turning the rows into model objects.
    257         """
--> 258         self._fetch_all()
    259         return iter(self._result_cache)
    260

/usr/local/etc/virtualenvs/gears/local/lib/python2.7/site-packages/django/db/models/query.pyc in _fetch_all(self)
   1072     def _fetch_all(self):
   1073         if self._result_cache is None:
-> 1074             self._result_cache = list(self.iterator())
   1075         if self._prefetch_related_lookups and not self._prefetch_done:
   1076             self._prefetch_related_objects()

/usr/local/etc/virtualenvs/gears/local/lib/python2.7/site-packages/salesforce/backend/query.pyc in iterator(self)
    235         """
    236         try:
--> 237             sql, params = SQLCompiler(self.query, connections[self.db], None).as_sql()
    238         except EmptyResultSet:
    239             raise StopIteration

[...]

PR #177, opened 2017-03-02
The test for fixed issue #175

hynekcer:


philchristensen:
Getting the error: "AttributeError: 'ForeignKey' object has no attribute 'related_model'". Not sure; do I need to add something to my test account?

https://travis-ci.org/django-salesforce/django-salesforce/builds/206851587

hynekcer:
Yes, it is in Django 1.7. My fifth commit in the development branch was to remove Django 1.7 from tests an README and then removed the related code in the next commit. Then is improved support for new Django versions etc. I don't develop for 1.7 any more.

philchristensen:
Oh, of course. It's late here, but I'll cherry-pick those commits tommorow.

PR #182, opened 2017-04-28
support django 1.11

tuky:
might need improvements here and there and a little rough, but for us it is working.

### DIFF
--- a/salesforce/backend/compiler.py
+++ b/salesforce/backend/compiler.py
@@ -446,7 +446,11 @@ class Range(models.lookups.Range):
         if connection.vendor == 'salesforce':
             lhs, lhs_params = self.process_lhs(qn, connection)
             rhs, rhs_params = self.process_rhs(qn, connection)
-            assert rhs == ['%s', '%s']
+            lhs = list(lhs) if isinstance(lhs, tuple) else lhs
+            lhs_params = list(lhs_params) if isinstance(lhs_params, tuple) else lhs_params
+            rhs = list(rhs) if isinstance(rhs, tuple) else rhs
+            rhs_params = list(rhs_params) if isinstance(rhs_params, tuple) else rhs_params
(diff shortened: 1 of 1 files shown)


philchristensen:
We really appreciate the submission! Unfortunately this isn't passing tests for me on dj111 and dj18: 

https://travis-ci.org/django-salesforce/django-salesforce/builds/226877895

hynekcer:
I also wrote a fix for Django 1.11 a few days before Easter. Then I tried to improve some feature in order to decide, which direction of development is better, to prevent not to announce something that must be redefined or removed soon. The pause during Easter and too much other work next week... A discussion would be probably useful if you or I would know on what it' working on and what burns also other people. I review it and accept probably very much.

tuky:
@hynekcer Do you think you could extract only the 1.11 compatibilty fixes and only PR those or include them here? I think it is important to get that right, before changing other fundamental things. Unfortunately I don't have the knowledge of django db backends and 1.11 seems to have some bigger changes, e.g. due to the new explicit sub-queries.

hynekcer:
Tuky@ thanks for your code, it is partly inspirative and it wold be very useful, but it has been solved still e.g. here https://github.com/hynekcer/django-salesforce/tree/tuky-fix-pr-182 As you see I accepted all your and Zee93 code except code that was reverted or fixed by you immediately. I will rename and reorganize my commits, but the result code passes all tests still. The only problem is in bulk update in non trivial cases, that is marked as expected failure now, if QUIET_KNOWN_BUGS is set.

hynekcer:
A new PR #183 with many your code is ready. Especially your solution with back-ported Iterator class is very appreciated. My original solution was "defensive" to can not modify or break anything working still, even that new features were emulated on old ones. Your solution looks safe. Thanks very much.

PR #183, opened 2017-04-29
Fixed for Django 1.11

hynekcer:
Based on parts from #182 (tuky - Tobias Krönke)
A new release is useful, even that it is like RC version (release candidatus) if used with  Django 1.11 and a warning about it is written.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -10,7 +10,7 @@ django-salesforce
 .. image:: https://img.shields.io/badge/Python-2.7.9%2B%2C%203.4%2C%203.5-brightgreen.svg
    :target: https://www.python.org/
 
-.. image:: https://img.shields.io/badge/Django-1.8.4%2B%2C%201.9%2C%201.10-blue.svg
+.. image:: https://img.shields.io/badge/Django-1.8.4%2B%2C%201.9%2C%201.10%2C%201.11-blue.svg
    :target: https://www.djangoproject.com/
 
 This library allows you to load and edit the objects in any Salesforce instance
@@ -19,10 +19,10 @@ for most uses. It works by integrating with the Django ORM, allowing access to
 the objects in your SFDC instance (Salesforce .com) as if they were in a
 traditional database.
 
-Python 2.7.9+, 3.4, 3.5, Django 1.8.4+, 1.9, 1.10 are supported with
+Python 2.7.9+, 3.4, 3.5, Django 1.8.4+, 1.9, 1.10, 1.11 are supported with
 some limitations on raw queries, values_list() and values() methods.
 
-Django 1.10 is currently supported without values(), values_list(), defer(),
+Django 1.10+ is currently supported without values(), values_list(), defer(),
 some raw() methods and without makemigrations. Full support of 1.10 is
 expected near the end of Q3 or beginning of Q4 in 2016.
 
--- a/README.rst
+++ b/README.rst
@@ -19,12 +19,11 @@ for most uses. It works by integrating with the Django ORM, allowing access to
 the objects in your SFDC instance (Salesforce .com) as if they were in a
 traditional database.
 
-Python 2.7.9+, 3.4, 3.5, Django 1.8.4+, 1.9, 1.10, 1.11 are supported with
+Python 2.7.9+, 3.4 to 3.6, Django 1.8.4+, 1.9, 1.10, 1.11 are supported with
 some limitations on raw queries, values_list() and values() methods.
 
-Django 1.10+ is currently supported without values(), values_list(), defer(),
-some raw() methods and without makemigrations. Full support of 1.10 is
-expected near the end of Q3 or beginning of Q4 in 2016.
+Django 1.10 and 1.10 is currently supported without values(), values_list(), defer(),
+some raw() methods.
 
 Pre-2.7.9 Python versions don't have the required TLS 1.1 support to use
 both production and sandbox Salesforce instances. PyPy may still work,
--- a/README.rst
+++ b/README.rst
@@ -22,8 +22,9 @@ traditional database.
 Python 2.7.9+, 3.4 to 3.6, Django 1.8.4+, 1.9, 1.10, 1.11 are supported with
 some limitations on raw queries, values_list() and values() methods.
 
-Django 1.10 and 1.10 is currently supported without values(), values_list(), defer(),
-some raw() methods.
+Django 1.10 and 1.11 is currently supported without values(), values_list(), defer(),
+some raw() methods. (All fixed in a development repository, waiting for review,
+consensus etc.)
 
 Pre-2.7.9 Python versions don't have the required TLS 1.1 support to use
 both production and sandbox Salesforce instances. PyPy may still work,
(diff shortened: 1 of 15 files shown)


README.rst
```diff
@@ -19,12 +19,11 @@ for most uses. It works by integrating with the Django ORM, allowing access to
...
 traditional database.
 
-Python 2.7.9+, 3.4, 3.5, Django 1.8.4+, 1.9, 1.10 are supported with
+Python 2.7.9+, 3.4 to 3.6, Django 1.8.4+, 1.9, 1.10, 1.11 are supported with
 some limitations on raw queries, values_list() and values() methods.
 
-Django 1.10 is currently supported without values(), values_list(), defer(),
-some raw() methods and without makemigrations. Full support of 1.10 is
-expected near the end of Q3 or beginning of Q4 in 2016.
+Django 1.10 and 1.10 is currently supported without values(), values_list(), defer(),
```
tuky: small typo: `1.10 and 1.10`

salesforce/backend/query.py
```diff
@@ -483,6 +513,12 @@ def execute(self, q, args=()):
             self.results = iter([])
 
     def execute_select(self, q, args):
+        if q == "SELECT django_migrations.app, django_migrations.name FROM django_migrations":
```
tuky: I never would have anticipated fixes like this. Do you have any documentation that I don't know? In general I can't find any docs on how to write a custom dajngo db backend, which is why I guess you have to live with lots of undocumented and "tricky" code.

hynekcer: You are right about documentation. Any help will be appreciated, even to track the project only and to review and early or constructive comments. My language skills are problematic when I described a complex "moving target" and other people can not understand it in README clearly. They respond too late or propose a fix by removing. Not too motivating. I started then to write more info to Wiki and even more info to [hynekcer sf wiki](https://github.com/hynekcer/django-salesforce/wiki/), where all original project topics are linked together and some independent topics (urls) are added. (My repository looks messy but the Wiki is a text for other people or a concept of this) I can write code or documentation, but usually it is very innefficient to try to do it synchronized. The objective is to write a pseudo-documentation during development, use it for discussion and later to be rewritten to documentation by other person. I have no ambition to start multi-page documentation, even that it would be useful to differentiate e.g. easy install :-) from the reference.
 
It even would not help me to have a documentation for writing custom db backend, because nobody reasonable from Django team would promise any interface for NoSQL databases. In the recent Django versions since 1.8 you can see a backend specific code not only in `django/db/backends/', but also in 'django/db/models/', e.g. many `as_mysql` methods. There is no clean way to write an additional backend without tricky code. I remember previous versions of MongoDB for Django, where it was even more tricky due to Mongo driver for patched Django, up to version 1.5.

"Tricky"? As little as possible. Example: Your code is by the standard recommended style, e.g. the try/except branch for new/old code. My experience is, when I must observe undocumented information and also must regularly remove obsoleted code after previous Django versions, that the explicit style ("if django version > something; else;) is better for tests, debugging and also to get more useful tracebacks in issues.

Migrations: First: Search the string "migrat" in README (4 lines in 3 paragraphs) and marginally also [Introspection and Special Attributes...](https://github.com/django-salesforce/django-salesforce/wiki/Introspection-and-Special-Attributes-of-Fields) (1 paragraph) ... (after you read) ...
Migrations are more and more important in Django, even the current "fake migrations" for Salesforce, if you for some reason want to use SalesforceModel with another database, e.g fast in memory tests, data copying etc. (see issues 56, 97, 98. 111, 167 that it is important.) This has been supported for a long time until Django 1.10 and fixed just now. My personal style to frequently run inspectdb on the current SFDC and run a filter to select important/unimportant fields for the final model. If SFDC should alternate with another db in development or maintenance tasks, then run makemigrations an migrate to get the same model in that database. If you can try it and explain it human way, do it please. :-)

hynekcer:
Tuky@ I wrote a very long comment below the code commented by you and rewrote it, so that the code line is  outdated, but hopefully my answer/wish is still interesting.

philchristensen:
I'm having a hard time getting Travis to have the correct Python versions installed, please bear with me.

philchristensen:
This is passing, it just looks like it's not. Thanks to @BanzaiMan for his patience dealing with all of us Pythonistas on travis-ci/travis-ci#4990

philchristensen:
I think I'll be releasing this as 0.7 in a day or two.

salesforce/backend/query.py
tuky: These are very interesting insights, thank you! Regarding faster tests / development with SQLite as a test backend for SF: we were using it heavily in the beginning, but came to the conclusion, that it is not a good replacement. SF itself implements so much logic, that cannot be easily "simulated" with a normal DB. This does not even include special `Flow` events.

hynekcer: I agree with you. I wrote the same point of view in some of these issues, if I remember. Fortunately the support makemigrations is marginal and not complicated.

If you are interested about "insights": My development repository, where all "problematic" methods are implemented again, is going to implement some parts of Python DB API and to be less "hacky" and less dependent on observing new Django internals. Such code is more laborious once, but should be easier repeatedly after new Django, because it depends more on documented API. I still hesitate about backward compatibility of that direction.

philchristensen:
Released: https://pypi.python.org/pypi/django-salesforce/0.7.0

tuky:
Thanks again for the release. I think, it lacks some incremented version numbers. Installing `0.7.0`, but `pip freeze` shows `0.6.9` afterwards:

```bash
(sdgsdgsd) tobias@flash:/tmp$ pip install django-salesforce==0.7.0
Collecting django-salesforce==0.7.0
Collecting requests>=2.0.0 (from django-salesforce==0.7.0)
  Using cached requests-2.14.2-py2.py3-none-any.whl
Collecting pytz>=2012c (from django-salesforce==0.7.0)
  Using cached pytz-2017.2-py2.py3-none-any.whl
Collecting django<1.11.99,>=1.8.4 (from django-salesforce==0.7.0)
  Using cached Django-1.11.1-py2.py3-none-any.whl
Collecting simplejson>=2.5.0 (from django-salesforce==0.7.0)
Installing collected packages: requests, pytz, django, simplejson, django-salesforce
Successfully installed django-1.11.1 django-salesforce-0.6.9 pytz-2017.2 requests-2.14.2 simplejson-3.10.0
(sdgsdgsd) tobias@flash:/tmp$ pip freeze | grep salesforce
django-salesforce==0.6.9
```

hynekcer:
Yes, the issue about old version number is #184. @philchristensen responded yesterday about it and it should be solved besides other things by PR 185 to 0.7.1.

(This package had used other release style for a long time, consistently derived from git tag number. The disadvantage was that the version number was not readable from the imported package and even not correctly readable by no conceivable preferable complicated method, under some circumstances. The new style is not routine yet overleaf. I usually never bump the version in order to leave the right of the last commit to Phil Christensen.)

tuky:
Yes thanks. Should have looked before ^^

Issue #174, opened 2017-02-15
HttpsReadTimeOutError


uditagarwal:
Thanks for this extremely useful plugin. It's working well for me.

However, I see a bit of an issue. If I use the plugin after about 12 hours, my first call to the SalesForce API to create a new entry always fails because of a TimeOut error. Subsequent requests pass without an issue, its just the first request after a long time that causes it to error.

Is there a way I can solve this?

hynekcer:
Yes, it is solved in some my old development branch. I will look at it and commit later. (busy)

hynekcer:
I remember, that a temporary solution before the fix was to run a cron job with a request every 119 minutes or more trequently.

My new solution uses a smart session and authentication management that prevents many similarly unfrequent problems, usual at client software for network services. Expect a beta quality code that could be worse eventually in some area until it is tested intensively, e.g. also by you. Don't expect anything before a month later, please.

uditagarwal:
Hey!

Thanks a lot for your response. Can you please link to the development branch so that I can see if i can use it on my project whenever you find the time?

I'll wait for your solution!

hynekcer:
It is  
`pip install -e git+https://github.com/hynekcer/django-salesforce.git@b92c5f2c2c97b3434d554fd5a90aaa5a4ce9134b#egg=django-salesforce`
(I refer to a fixed commit head because my development branch can be heavily rewritten whenever I write many fixing commits for a code that is still in the development branch and I move these commits to the right places and squash together.)

One important fix is "Timeout as tuple" (shorter for connect, longer for queries - data)

    SALESFORCE_QUERY_TIMEOUT = (4, 15)  # default (connect timeout, data timeout)

Other important fix is "Prevention SF account locking" if you e.g. must change a password and there are many requests accidentally before the password can be changed also on web server. Excessive number of failed login requests is banned by SF even if you don't specify a maximal number of allowed failed login.

Useful info about the branch is in a recent issue #173.

hynekcer:
This issue is rather improved ty timeout as tuple (a commit in #185). It requires requests 2.4 or higher.

A possible timeout, after the connection was inactive longer than 120 minutes, but the closing packet from SF was lost for some reason and the local part of the connection is still supposedly open, is still not solved. (It depends on big refactoring that can not be committed to master yet. This issue remains open.)

hynekcer:
Fixed

* Added: command ping_connection that automatic called after every
[...]

PR #185, opened 2017-05-15
Small fixes and a more granular timeout value as tuple

hynekcer:
The timeout tuple is related to #174


hynekcer:
* The most important change is the updated version number 0.7.1 in salesforce/__iit__.py and removing all refefences to the old VERSION file.
* README and demo configuration how to use two timeouts, a short for connecting and a longer for requests.

Other commits are important rather internally:
* "uid_version" in "test_helpers" to be possible to run a long batch of tests and to run an individual test on the same sandbox while debugging. Now I prefer to identify some test objects only by virtualenv name "py..-dj...", because it is more useful than the previous identifier by random numbers.
* Refactored escaping of SOQL arguments to the same resulta with cleaner code.
* The validity of README.rst can be checked by `tox -e docs`
* Introspection by inspectdb can be tested by: SLOW_TESTS=on tox

I'm still thinking about the word "quality" in the current warning "Support for Django 1.11 is still in pre-release quality. Test your app properly after upgrading."

philchristensen:
I'm still having issues with this coming up as 0.6.9 instead of 0.7.1.

PR #191, opened 2017-07-29
Fixed router bug with an alternate SF database. close #160

hynekcer:
(Maybe I add a commit here with tests and docs, but I want to show a solution that has been well tested manually. Wait same days with merge.)

PR #194, opened 2017-08-05
Fixed router bug with an alternate SF database. close #190

### LINKED ISSUE #190: Best way to run in environments without connectivity to Salesforce API?
This plugin is working great for us and we have an inquiry/documentation request. 

**Question:** What is the best way to run a django server that includes django-salesforce and no available Salesforce instance?

**Background:** We've been exploring is running our apps for development and CI/CD purposes when there is either no access to a Salesforce API or even no access to internet entirely (no active network interfaces except localhost). The dream is that we can run our apps and continue developing in these rare case we have no internet - like on a beach somewhere remote or on an airplane without internet on the way to getting to that beach.

**Possible Solutions?:** My understanding is that one option is to point to a local instance of SQLlite. Another option we've looked into is disabling the System Check Framework for the app where we use django-salesforce.

Currently if we disconnect from the internet and try to run our apps we get stuck with 'Performing system checks...' How do we move beyond this issue?

hynekcer:
* Fixed the referenced issue number PR #191
* Support for setting primary keys on objects created using bulk_create() with Django 1.10+ (This feature was supported only by Postgres, but it is much more important with SF than with other databases due to missing transactions)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -154,10 +154,11 @@ Advanced usage
 
 -  **Testing** - By default, tests will be run against the SFDC connection
    specified in settings.py, which will substantially increase testing time.
-   
+
    One way to speed this up is to change the SALESFORCE_DB_ALIAS to point to
    another DB connection (preferably SQLite) during testing using the
-   ``TEST_*`` settings variables. Django unit tests without SalesforceModel
+   ``TEST_*`` settings variables. Such simple tests can run without any network
+   access. Django unit tests without SalesforceModel
    are fast everytimes. Special read only fields that are updated only by SFDC
    e.g. ``last_modified_date`` need more parameters to be possible to save them
    into an alternate database, e.g. by ``auto_now=True`` or to play with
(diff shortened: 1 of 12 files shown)


philchristensen:
I'm getting an error on this test:

```
======================================================================
ERROR: test_bulk_create (salesforce.tests.test_integration.BasicSOQLRoTest)
Create two Contacts by one request in one command and find them.
----------------------------------------------------------------------
  File "/home/travis/build/django-salesforce/django-salesforce/salesforce/tests/test_integration.py", line 443, in test_bulk_create
    self.assertTrue(ret and all(x.id for x in ret))
  File "/home/travis/build/django-salesforce/django-salesforce/salesforce/tests/test_integration.py", line 443, in <genexpr>
    self.assertTrue(ret and all(x.id for x in ret))
AttributeError: 'Contact' object has no attribute 'id'
```

Unless I'm misreading it looks like it's only erroring on 3.5, which seemed weird.

hynekcer:
This was a backward compatibility with `SF_PK='Id'`.

PR #195, opened 2017-08-09
Fixed deserialize with app_label config. close #171

### LINKED ISSUE #171: app_label logic in prep_for_deserialize reversed?
https://github.com/django-salesforce/django-salesforce/blob/master/salesforce/backend/query.py#L147-L150

```
mod = model.__module__.split('.')
if(mod[-1] == 'models'):
    app_label = mod[-2]
elif(hasattr(model._meta, 'app_label')):
    app_label = getattr(model._meta, 'app_label')
```

If the user has specified `app_label` in the meta shouldn't that take precedence over the module path?

hynekcer:


hynekcer:
Excuse, I don' want to commit "Require 'salesforce' in installed apps", only the fix 171. (It is usually useful to check the presence of "salesforce" app in settings, but I don't want to do it in the the root this way.)

hynekcer:
The last commit is replaced in the next new PR.

PR #201, opened 2017-12-12
sensible version range

graingert:


philchristensen:
A little more detail would be appreciated. Why do you need this change?

graingert:
1.11.98 is going to be compatible with 1.11.99 and 1.11.100

Python versions are not limited to two figures per place

philchristensen:
eh. works for me.

hynekcer:
I can accept it eventually, but I don't agree that the new string is better.

The purpose for "Django>=x.x,<x.x.99" was that I wanted to exclude beta version and release candidates, because e.g. `1.12b1 < 1.12`. I'm sure of 99.9% that new major versions will not work without many fixes and anyone would usually only waste time if he tries it or even if he tries to fix all issues if he don't want to became a regular contributor. This a special consequence of the current Django-salesforce implementation and I don't know any similar package where must be an universal "package<version" before that version is released.

The highest Django patch release version was 1.4.22 in the 1.4 LTS feature release. The number 99 is very improbable because patch versions are usually released at the beginning of month for the period of mainstream support. Special security releases are much less frequent. Django-salesforce is usually upgraded to the new Django feature release before the tenth patch. This is a reserve one order.

Another reason was a historical incompatibility of beta version parsing and compare in package managers etc. like setuptools, distribute, distutils, pip, tox... Maybe it is now better.

The main reason why I want to reject this fix is a compatibility with version version numbers in [tox.ini](https://github.com/django-salesforce/django-salesforce/blob/master/tox.ini) file. We should change all or nothing.

@graingert What you mean about the word "sensible"?

graingert:
< 1.12 excludes beta versions.

philchristensen:
Oh is that why the last unittest run at travis failed? I'm trying to fix that today, so I haven't done any research.

Sorry, I guess I should have tested this on a local branch, it just seemed kind of small and inconsequential. I'm going to roll back this change for now.

PR #205, opened 2017-12-24
salesforce backend query fix for the case of filtering by a related_n…

nirkra:
…ame subfield

### DIFF
--- a/salesforce/backend/query.py
+++ b/salesforce/backend/query.py
@@ -142,7 +142,25 @@ def prep_for_deserialize(model, record, using, init_list=None):
     specified, then only these fields are processed.
     """
     from salesforce.backend import base
-    # TODO the parameter 'using' is not currently important.
+    if record['attributes'].get('type', None) != model._meta.db_table:
+        # this is for objects that were filtered using a reverse relationship field (OneToOne or ManyToMany)
+        # this is required because salesforce return hierarchical JSON, and the parent object is first.
+        # e.g. Dog(owner=Person.object.get(name="barperson",name="foodog").
+        # now we want to look for Persons that have a dog name foodog, in the case an owner can have more than one dog.
+        # now lets say the dog model looks like this:
+        # class Person(models.Model):
+        #   name = models.CharField(name=32)
+        # class Dog(models.Model):
+        #    owner = models.ForeignKey(Person, related_name="dogs")
+        #    name = models.CharField(max_length=32)
+        # now we want: list_of_names_of_owners_of_foodog = [i.name for i in Persons.object.filter(dogs__name="foodog")]
+        #
+        record.pop('attributes')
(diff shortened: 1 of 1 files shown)


philchristensen:
I keep coming back and looking at this, but I don't get what it's doing, or what bug it's fixing.

Could you provide a little more context?

hynekcer:
Yes, also a test case is useful because it is probable that we can re-implement this differently and some original usage would be broken.

SFDC doesn't support a custom unique field (if I know - checked one year ago). Therefore OneToOne relationship seems impossible.

hynekcer:
Maybe it can be accepted. It seems currently so fragile that I would leave it undocumented. What you mean by "OneToOne" in Salesforce?

salesforce/backend/query.py
```diff
@@ -142,7 +142,25 @@ def prep_for_deserialize(model, record, using, init_list=None):
...
+        # now lets say the dog model looks like this:
+        # class Person(models.Model):
+        #   name = models.CharField(name=32)
+        # class Dog(models.Model):
+        #    owner = models.ForeignKey(Person, related_name="dogs")
+        #    name = models.CharField(max_length=32)
+        # now we want: list_of_names_of_owners_of_foodog = [i.name for i in Persons.object.filter(dogs__name="foodog")]
+        #
+        record.pop('attributes')
+        if len(record.keys()) == 1 and record[list(record.keys())[0]].get('attributes', {}).get('type', None) == model._meta.db_table:
```
hynekcer: I will simplify it to:
```
if len(record.keys()) == 1 and list(record.values())[0].get('attributes', {}).get('type', None) == model._meta.db_table:
```

salesforce/backend/query.py
```diff
@@ -142,7 +142,25 @@ def prep_for_deserialize(model, record, using, init_list=None):
...
+            record = list(record.values())[0]
+            if record is None:
+                return None
```
hynekcer: This line is never reached because if `list(record.values())[0]` is None then also `record[list(record.keys())[0]]` on the line above was None and it failed with the error `AttributeError: 'NoneType' object has no attribute 'get'`. You could verify this bug by running test `test_integration.BasicSOQLRoTes.test_one_to_one_field`. The test for None must be before `.get('attributes',...)`.

hynekcer:
The class ApexEmailNotification is a special case where a unique foreign key exists. All such OneToOne fields exist only in a few system objects. I don't expect that anybody except Salesforce engineers use them in models and not at all that he needs easy advanced queries on them in Django.

hynekcer:
Here is a [fix for your patch](https://github.com/hynekcer/django-salesforce/commit/925ac607d156549f66b39ca84f931868bf29ed2f). I still need to write a new test.

philchristensen:
Closed in favor of #223

PR #216, opened 2018-12-07
Changes for Django 2.x compatibility

jbaack:
Start of changes for 2.x compatibility. This runs and passes my very basic manual tests for 2.1.3 as well as 1.11.


hynekcer:
It seems very good for a new development branch if say that you can run your tests.

It is very interesting, how it can work with your code. Can you please share an example? I see the exception is e.g.  
`AttributeError: type object 'Contact' has no attribute 'objects'`  
for all models. It doesn't work for me neither with Django 1.11 or Django 2.x. It is due to my old regression error that broke custom managers in the commit c3592579 from April 2015. That is revealed  combined with your line  
`sf_manager = manager.SalesforceManager()`. It works for me some way only without that line and without setting "base_manager_name" and I fixed also many issues, but I am interested in the alternative, how works for you.

jbaack:
I realized inspectdb was not working correctly - I was testing with a model generated from D1.11. One minor additional change was required - that is now in the PR. I also added a file testSFmodel.txt that is the large, full Salesforce database model.py file I am using to test. Other than that, I am just loading D2.1.3, adding django-salesforce and running the admin. I am using this in admin.py to add all models to the admin interface
register_omitted_classes(models)
Accounts, Contacts and other normal objects work fine.

hynekcer:
Thank you very much for the great information that Admin (/admin/) works still rudimentary for browsing the data. I usually try the admin reasonably not before many database tests starts to work.

This PR works some way, e.g. `Contact._base_manager.all()` that is used by admin or `Contact.sf_manager.all()`, but the most usual manager `objects` example `Contact.objects.all()` raises the exception above.

I renamed your `sf_manager` to `objects` in order to fix it. Everything important works now what worked in Django 1.11, only two small features had to be silenced in tests. Please try it. Why you used the new name sf_manager not objects? My current fixed version is [hynekcer/django20](https://github.com/hynekcer/django-salesforce/tree/django20).

I edited yesterday the previous comment. (The incorrect sentence is crossed out.) The truth is contrarily that  Django-salesforce does correctly the same as Django: If a custom manager is added with another name, then the model has no automatic attribute "objects". Your code is useful for debugging, but not compatible with most of standard projects.

jbaack:
Thanks. That code change was based on my interpretation of this - but I have no doubt it may not be complete so if your version is running better that is great news. I will give it a try.

https://github.com/django/django/commit/ed0ff913c648b16c4471fc9a9441d1ee48cb5420

hynekcer:
I accepted your first two commits by including them in my PR #217, but the entire PR must be rejected due to the last commit with a huge model. Thank you again for inspiration with the manager.

PR #217, opened 2018-12-12
Fixed for Django 2.1

hynekcer:
Django versions from 1.10 to 2.1 are supported in an almost equal quality, but still less than it was with Django 1.9. (version that is currently unsupported)

Included the first two commits from #216

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,8 @@ cache:
   directories:
     - $HOME/.pip-cache/
 before_install:
- - pyenv global system 3.5
- - if [[ $TOXENV = py35 && -f ~/virtualenv/python3.5/bin/activate ]]; then source ~/virtualenv/python3.5/bin/activate; fi
+ # - pyenv global system 3.5
+ # - if [[ $TOXENV = py35 && -f ~/virtualenv/python3.5/bin/activate ]]; then source ~/virtualenv/python3.5/bin/activate; fi
  - "pip install setuptools_git"
  - "pip install rstcheck"
 install:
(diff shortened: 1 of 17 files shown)


hynekcer:
I think that this PR is roughly all for Django 2.0 / 2.1 that could be done fast and that is considered relative stable, in order to release a new version (next week?)

philchristensen:
So, the Travis check above failed because of the continuing issues with making all the necessary Python versions available on Travis. I'll work on that independently.

But do we want to bother targeting Python 3.5 anymore? It seems everyone is finally getting onboard the Python 3 train around 3.6. Just a thought....

chromakey:
Ubuntu 16.04 LTS won't upgrade past python 3.5 with the official apt repos.  Not sure if that matters to you guys or not, but just thought I'd chime in with it.

philchristensen:
@chromakey thanks, that sounds like a good enough reason to me to keep testing on it. I’m sure this is a simple issue, i just haven’t taken a deeper look yet

hynekcer:
I found and I fixed now a bug in generator with StopIteration that was specific to Python 3.7. I had to start to check regularly also Python 3.7 therefore.
I test every last commit before push also with Python 3.5 and 3.7, but I temporarily replaced them in Travis by Python 3.6.

Please re-run Travis on the latest commit to see if the commit succeeded.

philchristensen:
@hynekcer If you have a moment and could pull/push the additional commit I added at django-salesforce/django-salesforce@0ee92187, that will update the Travis config to use Xenial and re-add support for 3.5

PR #220, opened 2019-01-08
Fixed support for Salesforce 45.0 Spring '19 and fixed inspectdb

hynekcer:
Some bugs in inspectdb are fixed now. The most important is caused by simplified sobject describe in the new Salesforce 45.0 Spring '19, currently known in Preview sandboxes. I think that it is a reason for a new version 0.8 now.

The version 45.0 works fine now even with API 45.0, but the compatible API 44.0 is configured in master until all instances migrate.


hynekcer:
I going to add something today.

hynekcer:
Please commit it. A new pull request will depend on it.

philchristensen:
Looks like a small issue with respect to the "id" field. My test account still uses "Id".

philchristensen:
I did merge in #223 because it was a small one, so it's probably also worth rebasing.

hynekcer:
All these commits are a part of a longer PR #222 that has been merged yesterday.

PR #222, opened 2019-02-05
Cumulative commit for v0.8

hynekcer:
This is all code. All tests passed. I wrote it as a cumulative commit.

0.8 Implemented a big part of Python DB API 2.0
-----------------------------------------------
* Standard DB API is emulated for all ``select`` commands, because it is
  finally easier and much more stable than to keep the old monkey patch
  style in new Django versions.

* Implemented linear rows cursor, that is expected by Django, like in other
  databases, not the cursor with rows like nested multi level dictionaries.

* Implemented bulk methods ``queryset.update()``, ``queryset.delete()``,
  ``SomeModel.objects.bulk_create([SomeModel(...),...])``.
  Currently only for 200 rows, in transactions with AllOrNone option.
   The queryset must contain a restriction. It can be overridden e.g.
   by ``.filter(pk__gt='')``, that is everytimes true.

* Function ``len(...)`` can not be applied on ``RawQuerySet`` because
  the extension method ``__len__`` has been removed from RawQuerySet.
  (The current Django  doesn't cache the results objects of that queryset
  and it had no advantage, on the contrary converting the raw query set
  by ``list(queryset)``  would require two full queries with all data,
  if ``__len__`` is present.)

* New error reporting. Some errors can became a warning, configurable to
  raise exception or not, e.g. if a deleted or updated row has been
  deleted recently. Prepared also to a custom error handler to can report
  more errors, if the block operations could be run without AllOrNone
  transaction later.

* Some backward compatibility in raw queries can be reimplemented in the next
  version 0.9 for Django 2.2, if it will be required. A better
  compatibility with the standard Django is probably more important.

---

The original is the branch [dev-v0.8](https://github.com/hynekcer/django-salesforce/tree/dev-v0.8), but I will rewrite the history, maybe after the version 0.8, to make nice logical commits where all tests pass after every commit or failed test are marked by "expectedFailureIf". I expect no final difference between the ends of branches after rewriting.


philchristensen:
We should focus on fixing #220 first, but there seems to be one or two small issues here.

philchristensen:
👍

philchristensen:
👍

PR #206, opened 2018-02-18
Fixed PR #205 filter by remote m2m objects

hynekcer:
All text for this fix is in my comments to #205
Please don't combine my and nirkra's code by "Squash and merge" button. I prefer now "Rebase and merge", which is currently a fast forward and merge.

PR #223, opened 2019-02-05
Fixed PR #206 by rebase against master

philchristensen:
Applying #206 commits against latest master with Travis fixes.


hynekcer:
That is obsoleted. It depends on a code that is removed by refactoring. It must / can be solved in the query compiler.

philchristensen:
Should I roll it back, then?

PR #225, opened 2019-02-07
Unoffical support for Django 2.2 alpha 1

hynekcer:
The testing of Django 2.2 pre-releases can start.

Known issues:
    * inspectdb does not work. (Workaround: The model can be prepared by an older Django, until the issue is fixed)


dhbahr:
@hynekcer @philchristensen do you guys  have a plan to create a 0.8 release?

philchristensen:
Sorry, that’s partly on me. I’ll get this branch reviewed and merged today. @hynekcer thoughts on a 0.8 release after that?

philchristensen:
I've reviewed the branch, and created a draft release for v0.8. I can push that whenever.

hynekcer:
I'm going to review today and add eight commits from February 8th, especially the fixed inspectdb.

hynekcer:
This PR is ready.

Maybe you update `.travis.yaml` to test also Django 2.2b:  
`tox -r -e py27-dj111,py36-dj20,py36-dj21,py35-dj110,py36-dj22b`  
or also something with Python 3.7.

@philchristensen Thanks for the release notes draft. I'm going to peruse the code today if anything especially useful should be noted more there.

hynekcer:
That were my last commits for version 0.8

Some info, that may be useful for release notes.

Rows in raw queries are now lists by default, not dictionaries.

More complicated querysets with subqueries are supported. They are executed by one request now.

Select related is supported. It works also with minimal number of requests.

Better support for more numeric fields types:

FloatField type added, because it is known that a float number had been much more useful since Python 2.7 also for currency values storage than it was in previous Python versions: "[rounding problems solved](https://stackoverflow.com/a/35117668/448474)" thanks to modern reversible conversion str -> float -> str, without loss of accuracy and without superfluous inaccurate decimal places.

A DecimalField with zero decimal places is now `Decimal('0')` e.g. instead of previous `Decimal('0.0')`.

BigIntegerField - That type is not important for Salesforce alone, but if an object should be used with more database backends.

----

Issues:
- Dynamic fields are broken now, but they have been never officially announced. 
- USE_TZ = True  is necessary

These are not so important to wait for them now.

Issue #231, opened 2019-05-14
Edge case when testing against local database


ElliotSalisbury:
Hi,

Firstly, I love the work done on this, it's a really powerful tool.

I'm running into a problem when testing a salesforce database locally.
I believe the tables in the local database has been generated with an integer type for the ID field

For example:
```
a = Account(id="abc123abc123abc", ...)
a.save()
```
Would fail, as id has a type mismatch, and is expecting an integer

I believe this is because the `SalesforceAutoField`, as given to the id field in the `SalesforceModel` has not overriden the `get_internal_type()` method, and is defaulting to the integer type, of the parent class `AutoField`

Doing the following fixed my errors, but I wanted to check with you all first to see if I'm doing something wrong here. or if this should be pull requested?

```
class SalesforceAutoField(fields.AutoField):
    """
    An AutoField that works with Salesforce primary keys.

    It is used by SalesforceModel as a custom primary key. It doesn't convert
    its value to int.
    """
    description = _("Text")

    def get_internal_type(self):
        return "TextField"

    default_error_messages = {
        'invalid': _('This value must be a valid Salesforce ID.'),
    }

    def to_python(self, value):
        ...
```

Thanks
Elliot

philchristensen:
So, you wouldn’t ever really set that parameter in normal usage, given that it’s an auto-generated field.

Are you attempting to essentially pre-seed the DB with objects that have deterministic primary keys? It’s possible that we’ve never tested in exactly this fashion, so I’ll have to defer to @hynekcer as to whether this is a bug or a known limitation.

When you apply your patch, does it pass the django-salesforce test suite?

ElliotSalisbury:
I was writing some tests using a Factory to populate a local version of the salesforce db, something like this:

```
class SFAccountFactory(factory.django.DjangoModelFactory):
    class Meta:
        model = 'eiq_salesforce.Account'

    name = Faker('company')
    description = Faker('text')
    website = Faker('url')
```

But I have some code elsewhere that allows users to search for SF objects by ID, and it ensures that the ID is either 15 or 18 characters long, which these autogenerated IDs were not. So it was raising errors, and my tests could not pass.

So then I added an id field to the factory, to make it for sure 18 chars long
```
def get_standardized_sf_id(sf_id):
    # ensure ID is salesforce 18 char case insensitive id
    # https://help.salesforce.com/articleView?id=000274977&type=1&language=en_US

    suffix = ""
    for i in range(3):
[...]

PR #232, opened 2019-05-15
Model ids are now stored as Text when generating the tables of a local database

ElliotSalisbury:
From Issue #231

I'm running into a problem when testing a salesforce database locally.
I believe the tables in the local database has been generated with an integer type for the ID field

For example:

```
a = Account(id="abc123abc123abc", ...)
a.save()
```
Would fail, as id has a type mismatch, and is expecting an integer

I was writing some tests using a Factory to populate the local db, something like this:
```
class SFAccountFactory(factory.django.DjangoModelFactory):
    class Meta:
        model = 'eiq_salesforce.Account'

    name = Faker('company')
    description = Faker('text')
    website = Faker('url')
```
But I have some code elsewhere that allows users to search for SF objects by ID, and it ensures that the ID is either 15 or 18 characters long, which these autogenerated IDs were not. So it was raising errors, and my tests could not pass.

So then I added an id field to the factory, to make it for sure 18 chars long
```
def get_standardized_sf_id(sf_id):
    # ensure ID is salesforce 18 char case insensitive id
    # https://help.salesforce.com/articleView?id=000274977&type=1&language=en_US

    suffix = ""
    for i in range(3):
        loop = 0
        for position in range(5):
            current = sf_id[i * 5 + position]
            if current.isupper():
                loop += 1 << position
        suffix += "ABCDEFGHIJKLMNOPQRSTUVWXYZ012345"[loop]
    sf_id = sf_id[:15] + suffix
    return sf_id


class SFAccountFactory(factory.django.DjangoModelFactory):
    id = factory.Sequence(lambda n: get_standardized_sf_id('%015d' % n))
    ...
```

`get_standardized_sf_id` just ensures that the ID is following the same algorithm salesforce  uses to make the IDs case insensitive.

This then started throwing errors because it was trying to assign a string to an integer field in the database.

I believe this is because the SalesforceAutoField, as given to the id field in the SalesforceModel has not overriden the get_internal_type() method, and is defaulting to the integer type, of the parent class AutoField

```
class SalesforceAutoField(fields.AutoField):
    """
    An AutoField that works with Salesforce primary keys.

    It is used by SalesforceModel as a custom primary key. It doesn't convert
    its value to int.
    """
    description = _("Text")

    def get_internal_type(self):
        return "TextField"

    default_error_messages = {
        'invalid': _('This value must be a valid Salesforce ID.'),
    }

    def to_python(self, value):
        ...
```

I was not able to run the tests myself, I do not have access to a sandbox api.
Was hoping to use travis here.

### DIFF
--- a/salesforce/fields.py
+++ b/salesforce/fields.py
@@ -46,6 +46,9 @@ class SalesforceAutoField(fields.AutoField):
     """
     description = _("Text")
 
+    def get_internal_type(self):
+        return "TextField"
+
     default_error_messages = {
         'invalid': _('This value must be a valid Salesforce ID.'),
     }


philchristensen:
I think you're pretty close, but we're getting a few failures. I thinlk @hynekcer should be able to help us with these when he has some time.

[Test that the light weigh AutoField is compatible in all Django ver.](https://travis-ci.org/django-salesforce/django-salesforce/builds/533039754#L617):
`AssertionError: 'TextField' != 'AutoField'`

[test_capitalized_id (tests.test_migrate.tests.CompatibilityTest)](https://travis-ci.org/django-salesforce/django-salesforce/builds/533039754#L699):
`IntegrityError: NOT NULL constraint failed: Contact.Id`

ElliotSalisbury:
Actually, I see this was sorta tangentially discussed on #90

Since there are tests that ensure that `get_internal_type` is always 'AutoField' for backwards compatibility
Perhaps something like what @hynekcer said at the end of #90 is possible:
https://github.com/django-salesforce/django-salesforce/issues/90#issuecomment-71376728

We could check if the connection to the db is to a local backend or not.
and if it is, then we serialize the IDs from integers to the base 62 string as defined by salesforce, and back again.
So in the local DB they're integers, but in the python objects they're the 15/18 character strings?

philchristensen:
That seems like a possibility. My read on that comment is we didn’t have complete control of the default manager until after version 1.8, but support was dropped for that version in our last release, so now we should be good to go.

hynekcer:
I explained the general reason why I close this PR in the original issue #231. (It breaks the auto primary keys on the default database. No test exist for it currently.) The two failed tests are marginal, compared to the main problem.

The context of #90 has changed. More projects existed that used a custom patched Django to work with NoSQL databases. Django primary keys are fixed now in all supported versions, but if I know, only django-salesforce is active among these projects. If I remember the old code of Mongo, I think that these projects could be updated and could be maintained easier with new Django versions, but the reality is that they did not endure the crisis and django-salesforce is probably alone. I would never expect a patch in Django for one project. I don't like also to besmear a core functionality here by anything potentially problematic. A descendant of SalesforceModel is possible, initially in your code and if it works for you it can be accepted later to this project as an optional class.

I'm closing this for now.

Issue #226, opened 2019-03-15
Salesforce router causing issues with migrations


pbriet:
Hi,

Let's start by thanking everyone working on this project : great work!

This is my first issue with the library : i'm doing some completely unrelated work in an other database (MySQL), including renaming some models.
When I apply the migrations, I have a crash. It comes from the Salesforce router

```
  File "python3.5/site-packages/django/db/utils.py", line 291, in allow_migrate
    allow = method(db, app_label, **hints)
  File "python3.5/site-packages/salesforce/router.py", line 65, in allow_migrate
    model = apps.get_model(app_label, model_name)
  File "python3.5/site-packages/django/apps/registry.py", line 207, in get_model
    return app_config.get_model(model_name, require_ready=require_ready)
  File "python3.5/site-packages/django/apps/config.py", line 170, in get_model
    "App '%s' doesn't have a '%s' model." % (self.label, model_name))
LookupError: App 'XXX' doesn't have a 'YYYY' model.
```

Not sure what's the purpose of "allow_migrate" precisely. But what's surprising is that even by calling `./manage.py migrate --database default`, the issue remains  ("default" being my SQL database)

The only solution I've come up with is to remove the salesforce router from settings.py dynamically when applying migrations...

Any idea what could cause such issue?

Thanks

pbriet:
Django  2.1.7
django-salesforce  0.8

hynekcer:
I confirm a bug, that the ModelRouter fails if a migration is applied that deletes or renames a table, even in a normal database. It was a regression by version 0.8. I'm writing a PR with a fix. Thanks for reporting.

The purpose of `allow_migrate` method is to create the right tables in the appropriate database (salesforce or non salesforce).

---
If the table has been deleted yet in models then it is more complicated (impossible without guess to determine the right database. It is possible even after the fix, if a table is created in Salesforce and recorded to migrations and dropped and recorded the migration again, then another bug is possible. It is not serious: If a SalesforceModel table is deleted then it can be fixed by editing or deleting the Salesforce migrations history. It is acceptable because Salesforce migrations are used only for faster tests of Salesforce models on a non-Salesforce database.

pbriet:
Thanks for your feedback.

hynekcer:
[...]

PR #233, opened 2019-05-18
Version 0.8.1

hynekcer:
I want to release a new version 0.8.1, mainly in order to fix the regression #226, that was reported recently again.
The best day will be on Monday, May 20th, that date is in the changelog.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,6 +24,8 @@ experimental.
 
 * Add a new more clear class SalesforceAuthError report instead of LookupError.
 
+* Fix select_related() with a subquery by children objects. #234
+
 
 [0.8] 2019-03-06
 ----------------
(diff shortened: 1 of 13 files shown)


philchristensen:
I'm ready to release this today if it's still good to go.

hynekcer:
Yes, I think that the regression #226 in 0.8 reported last week again is a reason for a new version. I waited for time for writing tests, but it is better to not wait more. Probably spelling should be fixed in CHANGELOG or release notes draft please.

philchristensen:
Haven't forgotten about this, just haven't been able to get to it yet...

hynekcer:
Added a fix for #234. It is a safe change. Only the command "query_topology" is moved to the right place after `pre_sql_setup()` where the query mapping is finalized.

PR #237, opened 2019-06-05
Ping_connection and much more

### LINKED ISSUE #174: HttpsReadTimeOutError

### LINKED ISSUE #231: Edge case when testing against local database

hynekcer:
* Fixed: filter for objects with no children rows (missing test)  (e02de47)

* Added: lookup ``.filter(...__not_in=subquery)``' and 'not_eq',
  because of unsupported ``.exclude(...__in=subquery)``.  (988110c, 9dfb2de)

* Added: command ``ping_connection`` that automatic called after every
  longer inactivity to minimize timeouts. Fixed #174  (92fb900, 0117367)

* Fixed: ``makemigrations`` works now also without db_table name e.g. for
  simple standard objects.  (62245e5)

* Fixed: bug ``.using('salesforce')`` in ``default`` database queryset.  (dd00e5e)

* Added: ``salesforce.models_extend`` module with SalesforceModel with
  varchar primary key that works also with ``default`` databases.
  Fixed methods for it: save() and ``bulk_create()`` to can create a new
  pk or to copy an object exactly. Fixed #231 (4d042e5, 79603b7)

* Fixed: test setUpClass to can run tests on an empty Salesforce database.  (95ead69)


philchristensen:
@hynekcer that's a false failure in the Travis tests because of some kind of issue with migrations or something. I'm not really sure what to do to clear it, though...

hynekcer:
Hmm, a configurable SF_PK requires to use a corresponding migration. I removed the concrete migration and wrote a `makemigrations` command instead.

By the way, it can't be fixed by adding a new migration because:
> Renaming the 'id' column while in a transaction is not supported on SQLite < 3.26 because it would break referential integrity. Try adding `atomic = False` to the Migration class.

philchristensen:
That did the trick. Everything looks good to me now.

PR #238, opened 2019-09-17
dynamic auth makemigration issue

bryancolligan:
When using Dynamic Authentication and trying make a migration, a the dynamic auth functionality crashes the migration program without failing gracefully.

This PR allows for a graceful dynamic auth failure.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ tests/inspectdb/models.py
 salesforce/packages/
 venv
 tests/x_*/
+.idea/
--- a/salesforce/auth.py
+++ b/salesforce/auth.py
@@ -124,7 +124,7 @@ class SalesforceAuth(AuthBase):
 
     def __call__(self, r):
         """Standard auth hook on the "requests" request r"""
-        access_token = self.get_auth()['access_token']
+        access_token = self.get_auth().get('access_token', '')
         r.headers['Authorization'] = 'OAuth %s' % access_token
         return r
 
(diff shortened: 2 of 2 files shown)


philchristensen:
LOL, thanks for the PR anyway. I guess it wasn’t actually a bug?

hynekcer:
Thanks for a report about that special case. This should be fixed primarily by a documentation, however an exception should remain raised. 

The command `makemigrations` requires a working database connection to all databases. That is a Django requirement. It is indispensable. It is very very similar to the command `inspectdb` that need run also only on a development machine and it also needs to introspect a list of all tables in the database. Therefore lets have some local_settings with a static user account for management commands on a development machine, even with a configured individual dynamic auth used while running http requests.

The command `migrate` works correctly with a plain dynamic auth, because it works exactly on one database, usually the "default" (non-salesforce database) if nothing specified. Nobody would run an explicit `migrate --database=salesforce`. The command `migrate` is important also in production.

The list of all tables is necessary in `makemigrations` because it tries to check a presence of table `django_migratons` in that list:

    # file: django/db/migrations/recorder.py
    self.Migration._meta.db_table in self.connection.introspection.table_names(self.connection.cursor())

The presence of "django_migrations" table is simulated by a fake answer, because it is not important in SFDC databases. The list of all tables can't be faked, because it is also frequently a justified query.

hynekcer:
Updated wiki about it: https://github.com/django-salesforce/django-salesforce/wiki/Experimental-Features#user-content-dynamic-authorization

PR #239, opened 2019-09-20
Support for Salesforce 47.0 Winter '20 and Django 3.0

hynekcer:
I suggest a new release soon.

this PR:

- Fix a small bug with Salesforce Winter '20 (Currently present only in preview sanboxes, but coming very soon) It is quite an important bug, but the impact is only for a `BusinessHours` object if it would be updated by Django. It is the only object where the `TimeField` is used or can be used.

- Fix django-salesforce for Django 3.0 (currently 3.0 alpha 1) Problems were only with `insert` command. It is as well the only important change related to database in Release notes 3.0.

I expect only documentation edit and release notes before the new release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,4 +24,4 @@ install:
 script:
   - rstcheck README.rst
   - echo -e "SF_PK = 'Id'" > salesforce/testrunner/local_settings.py
-  - tox -r -e py36-dj30a,py27-dj111,py36-dj20,py36-dj21,py35-dj110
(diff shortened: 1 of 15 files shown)


philchristensen:
Let me know when you feel happy with everything, and I'll release v0.8.2.

hynekcer:
I have committed a change log now. The new version can be e.g. 0.9

When I see high jumps between recent Django versions I would probably suggest a version equal to the highest supported Django 3.x version next year, especially if we drop Python 2.7 for some reason later, after end of life Django 1.11 in April 2020.

Issue #247, opened 2019-12-23
Not compatible with django-debug-toolbar


l-kotzur:
Hey there,

first thanks for the great package. 

In order to avoid future headache of other developers, I just want to point out that django-salesforce was not compatible with the [django-debug-toolbar](https://django-debug-toolbar.readthedocs.io/en/latest/installation.html) for me. 

I got following error, running them together in a local environment:
```
 ...
  File ".../lib/python3.7/site-packages/django/db/models/base.py", line 741, in save
    force_update=force_update, update_fields=update_fields)

  File ".../lib/python3.7/site-packages/django/db/models/base.py", line 779, in save_base
    force_update, using, update_fields,

  File ".../lib/python3.7/site-packages/django/db/models/base.py", line 870, in _save_table
    result = self._do_insert(cls._base_manager, using, fields, update_pk, raw)

  File ".../lib/python3.7/site-packages/django/db/models/base.py", line 908, in _do_insert
    using=using, raw=raw)

  File ".../lib/python3.7/site-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)

  File ".../lib/python3.7/site-packages/django/db/models/query.py", line 1186, in _insert
    return query.get_compiler(using=using).execute_sql(return_id)

  File ".../lib/python3.7/site-packages/salesforce/backend/compiler.py", line 434, in execute_sql
    cursor.execute(sql, params)

  File ".../lib/python3.7/site-packages/debug_toolbar/panels/sql/tracking.py", line 192, in execute
    return self._record(self.cursor.execute, sql, params)

  File ".../lib/python3.7/site-packages/debug_toolbar/panels/sql/tracking.py", line 126, in _record
    return method(sql, params)

  File ".../lib/python3.7/site-packages/salesforce/backend/utils.py", line 181, in execute
    self.execute_select(q, args)

  File ".../lib/python3.7/site-packages/salesforce/backend/utils.py", line 250, in execute_select
    self.cursor.execute(soql, args, query_all=is_query_all)

  File ".../lib/python3.7/site-packages/salesforce/dbapi/driver.py", line 533, in execute
    raise ProgrammingError("Unexpected command '{}'".format(sqltype))

salesforce.dbapi.exceptions.ProgrammingError: Unexpected command 'INSERT'
```
Deactivating the django-debug-toolbar solved this problem.

[...]

PR #248, opened 2019-12-30
Debug toolbar

hynekcer:
I added three additional commit to easier reproduce the remaining problem issue #247 
after the older two commits where I fixed what could be reproduced.

* Updaded test environment (for Travis)
* Test insert, update and delete with django-debug-toolbar (can be used also for manual testing)
* Implemented EXPLAIN command (usable in django-debug-toolbar)

The EXPLAIN command is really a maximal possible support for Django debug toolbar.
See [Salesforce documentation about their EXPLAN command](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_query_explain.htm)

I will crate a new branch until the issue is fixed enough. (I found also some smaller issues by manual testing in example admin on a huge raw model from inspectdb.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,4 +24,4 @@ install:
 script:
   - rstcheck README.rst
   - echo -e "SF_PK = 'Id'" > salesforce/testrunner/local_settings.py
-  - tox -r -e py36-dj30,py27-dj111,py36-dj20,py36-dj21,py35-dj110
+  - tox -r -e py36-dj30,py27-dj111,py36-dj20,py36-dj21,py35-dj110,debug_toolbar
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,4 +24,4 @@ install:
 script:
   - rstcheck README.rst
   - echo -e "SF_PK = 'Id'" > salesforce/testrunner/local_settings.py
-  - tox -r -e py36-dj30c,py27-dj111,py36-dj20,py36-dj21,py35-dj110
+  - tox -r -e py36-dj30,py27-dj111,py36-dj20,py36-dj21,py35-dj110
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,4 +24,4 @@ install:
 script:
   - rstcheck README.rst
   - echo -e "SF_PK = 'Id'" > salesforce/testrunner/local_settings.py
-  - tox -r -e py36-dj30,py27-dj111,py36-dj20,py36-dj21,py35-dj110,debug_toolbar
+  - tox -r -e py36-dj30,py27-dj111,py36-dj20,py36-dj21,py35-dj22,debug_toolbar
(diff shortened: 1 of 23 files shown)


hynekcer:
Merged to a  branch `debug-toolbar`.

hynekcer:
* Fix: Support for Django Debug Toolbar - including EXPLAIN command

* Fix: Export more attributes by inspectdb for ForeignKey similarly to other fields.
     ("verbose_name", "help_text" and "default=DEFAULTED_ON_CREATE")
* Fix: Don't export DEFAULTED_ON_CREATE excessively for not createable fields.
* Fix: Error handling in bulk delete()

* Add: Support for the current snapshot of Django 3.1-alpha
* Remove: Support for Django 1.10 (incompatible with 3.1)
* Remove: Support for Python 3.4 (not tested any more)

I committed also this directly to a branch in origin because I had committed the first fix #247for [l-kotzur](https://github.com/l-kotzur) this way also and these commits are a side product.

Please review it and decide, eventually highlight some points for a future release draft or rewrite the news in Changelog..

---

I don't expect anything to pull until the end of January. Then probably a code for developer login by [SFDX](https://developer.salesforce.com/tools/sfdxcli) without a saved pasword and a very simple migration to can add some objects and fields into vanilla sandbox - like the old syncdb. This will improve tests and solve e.g. the issue # 62. I want also to write a fix for # 213 and # 221. Then I would like to release v1.0 with all these parts or some of them. A question is also if it should be the last version with Python 2.7 and Django 1.11, 2.0, 2.1 or these versions are somehow useful while compatibility is still easily sustainable.

philchristensen:
I'm not sure why the branch you pushed didn't trigger the Travis build, but my push did. Anyways, I ran that and everything looks fine to me.

Since we're on the cusp of dropping Python 2.7 support, I think we should do so before releasing 1.0. I don't think it's unreasonable to say that from 1.0 on we're Python 3 only.

It seems like we might want to do one last v0.9.x release while still supporting 2.7.

As for release notes, we can focus on version support updates (Python, Django, SFDC), and any additional features you feel are important to call out. I'll start a draft release shortly.

hynekcer:
Hmm it was not triggered because I wrote about pushing into an origin branch, but I forgot it or though the the pull request is good enough.

OK. The new features are mainly for developers and it doesn't make sense for Python 2.7. A simple bug-fix version 0.9.x will be released first, probably at the end of month before dropping Python 2.7.

I like to jump to version 3.1 in August when Django 3.1 will be supported. I think that it will be practical for many people. Support for a new Django version is the most important requirement of course. It seems that people never update Django-Salesforce without updating also Django.

philchristensen:
Oh right, I remember we’d start tracking the Django version. Sounds good to me.

hynekcer:
About SFDC version: The current code works with the new SFDC 48.0 Spring '20. A default setting API_VERSION = '48.0' will be marginal. It will add only some new internal objects and fields of SFDC. It can't be applied before 15 February 2020 (date of the latest Spring '20 release).

PR #250, opened 2020-03-06
Support local db with custom fields out-of-the-box

stefanfoulis:
Custom fields added in salesforce have the `custom=True` parameter on the field in django.
Because of how salesforce works regarding custom fields the name will get the `__c` suffix in the salesforce database (see https://github.com/django-salesforce/django-salesforce/blob/v0.9/salesforce/fields.py#L132). 

e.g when your salesforce model has a custom `language` field the `Account` model (other fields omitted for clarity):

```python
class Account(salesforce.models.Model):
        name = models.CharField(max_length=255, verbose_name="Account Name")
        language = models.CharField(
            custom=True,
            max_length=255,
            default=models.DEFAULTED_ON_CREATE,
            choices=[
                ("German", "German"),
                ("French", "French"),
                ("English", "English"),
                ("Italian", "Italian"),
                ("Spanish", "Spanish"),
            ],
        )
```

In salesforce the `name` field will be called `name`. But the `language` field will be called `language__c`.

When debugging locally using the local database rather than a real connection to salesforce we run into a problem though. Creating and applying the django migration the fields in the database end up not having this suffix. So when trying to access such a field we'd get an error from the db saying that the field does not exist.

```
  File "/virtualenv/lib/python3.6/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.ProgrammingError: column Account.Language__c does not exist
LINE 1: ...ount"."Description", "Account"."CurrencyIsoCode", "Account"....
                                                             ^
HINT:  Perhaps you meant to reference the column "Account.Language".
```

The problem is that the django migrations don't know anything about the `custom=True` field parameter, so at migration time it is unknown and assumed to be false and thus the `suffix` is not added to the fieldname.

The current workaround is to manually define `db_column ` on all such fields so that it works both on salesforce and in the local db. With this PR that will no longer be necessary.

The workaround in the current docs:

> Migrations - Migrations can be used for an alternate test database with SalesforceModel. Then all tables must have Meta options db_table and fields must have option db_column, which can be got by inspectdb with --verbosity=2. Models exported by introspection inspectdb do not specify the option managed because the default value True is considered safe. (Migrations in SFDC are not supported. If anything would be implemented after all, only explicitly clearly selected fields and models could be migrated in explicitly labeled SFDC databases. Consequently, the setting managed = True is related only to an alternate non SFDC database configured by SALESFORCE_DB_ALIAS.)

**This PR adds** a `deconstruct()` method on the salesforce fields to correctly store the `custom` field parameter in migrations.
This allows getting matching column names in the local db without having to manually explicitly set `db_column` on all custom fields if one wants to use the local db.
The migration machinery will automatically pick up on the change and will auto-generate valid migrations.

Impact/Upgrade path for existing users:
- User who are not using the local db (yet): no impact
- User who are using the local db but did not have any custom fields: no impact
- Users who previously used the local db by overriding `db_column`: no impact. They could remove the now unnecessary `db_column` parameters and run `manage.py makemigrations myapp` and `manage.py migrate` to catch up.
- Users who previously are using or tried to use the local db and either didn't notice the error or did and left the migration in the code anyway: `manage.py makemigrations myapp` will create a correct migration, even detecting the changed column names and rename them at db level. Then  `manage.py makemigrations migrate`.

Still needs tests and docs, but I wanted to get your eyes on it first to find our whether you are interested in merging this before I put in more work.

### DIFF
--- a/salesforce/fields.py
+++ b/salesforce/fields.py
@@ -102,6 +102,9 @@ class SfField(models.Field):
     """
     def __init__(self, *args, **kwargs):
         self.sf_read_only = kwargs.pop('sf_read_only', 0)
+        if 'custom' in kwargs:
+            # later used in self.deconstruct()
+            self._original_sf_custom = kwargs['custom']
         self.sf_custom = kwargs.pop('custom', None)
         self.sf_namespace = ''
         super(SfField, self).__init__(*args, **kwargs)
@@ -144,6 +147,24 @@ class SfField(models.Field):
         if self.sf_custom and '__' in cls._meta.db_table[:-3]:
             self.sf_namespace = cls._meta.db_table.split('__')[0] + '__'
         self.set_attributes_from_name(name)
+
+    def deconstruct(self):
+        deconstructed = super().deconstruct()
+        # Add the salesforce specific sf_custom param so it can be picked up by
+        # django migrations and have the same db field names in the local
+        # debug db as on salesforce.
+        keywords = deconstructed[3]
+        if hasattr(self, '_original_sf_custom'):
+            keywords["custom"] = self._original_sf_custom
+
+        # We are not handling sf_namespace here, because it currently can't be
+        # passed in to __init__ anyway. If a field subclass overrides
+        # sf_namespace the migration machinery can pick it up from there.
+
+        # We are not handling sf_read_only here because it can't be mapped
+        # directly to a django equivalent for the migrations and local db
+        # anyway.
+        return deconstructed
 
 # pylint:disable=unnecessary-pass,too-many-ancestors
 


philchristensen:
I’m not entirely clear what this “custom” parameter is for, could you elaborate?

stefanfoulis:
@philchristensen I added much more detail the the PR description ⬆️. And discovered that currently it is still necessary to set `db_table` in `Meta` on custom tables. I'll dig a bit deeper to see if that can also be avoided.

stefanfoulis:
Nevermind the stuff I was going on about still having to set `Meta.db_table`. Looks like that is by-design necessary for all models (custom or built-in), so not an issue. Sorry for the noise.

hynekcer:
I'm not sure that I understand the extent your problem:

> So when trying to access such a field we'd get an error

If you never use a raw SQL and you use only django field names, not internal field names then it probably works. If not then please write an example. (If a `db_column` is used in SQL than it must be hard coded also in the model.)

---

The parameter "`custom`" in fields and "custom" meta in models is now not fully obsoleted, but it is much less useful than in django-salesforce 0.4 (July 2014) when it was introduced. The idea was that most of builtin fields database names can be derived correctly from Django field names. The same is possible with simple custom field names easily. This seemed more Pythonic (don't repeat yourself) It was two months before Django 1.7, that introduced native migrations into Django. Support for local databases was added later also. That increased the preference `db_column` and preference of " Explicit is better than implicit."

I implemented two times a code for adding a custom model and custom fields to a Salesforce database schema (once by tooling API and once by SFDX), but it was never good enough for deployment (e.g. to add also write permissions and to can fix an undefined state if something failed in the middle database operations). It is not easy to write automated tests for custom fields that are widely reproducible in more Salesforce instances if we can't create that fields programmatically. I intentionally didn't add that stuff into migrations until these migrations will be useful in Salesforce. It can not be thoughtfully enough implemented. (If I remember, some migrations between South and Django 1.7 and 1.8 with normal databases were inconsistent, because the implementation in Django was not stable. Some older migrations had to be deleted and recreated by a new Django. That I want to prevent in django-salesforce)

I can not say generally that a table name "MyTable__c" is a better name for local databases than a default convention "myapp_mytable", because around ten standard Django tables are also present in a local database mixed with Salesforce tables. Imagine that django-salesforce can support simple migrations some day. Then normal simple django application could be deployed to salesforce and it will be useful to mangle only salesforce names and not the original non-salesforce names.

---

I infer this idea: **Support some salesforce parameters in `deconstruct()` like `custom=...` when migrations start be supported in Salesforce instances.**

Maybe the parameter `db_column` can be exported by inspectdb by default now without need to specify the verbosity. The minimalist db_column is not for the mainstream use.

stefanfoulis:
Ah, I see. So the `custom=True` is really more a relic of the past and the recommended way going forward would be to always define `db_column` on "custom" fields?
Then going forward we'd rather deprecate the `custom` parameter somehow. And make sure the introspection management command also does not output any code with `custom=True` anymore. 🤔

stefanfoulis:
I'll close this PR, as I don't think it aligns with the strategy of the project.

@hynekcer / @philchristensen would you agree that we should deprecate and eventually phase out `custom=True` though?

hynekcer:
It can be usually hidden in a new inspectdb because I agree that it is currently unclear, but it will remain optional. By the way, I worked yesterday again on a code related syncdb on Salesforce cloud.

hynekcer:
The `custom=` should be not a part of migrations from my point of view. It is not a database change. It is sometimes an alternative to `db_column` like a string in double and single quotes in Python are internally the same and both are useful.

Issue #236, opened 2019-06-04
Implement .bulk_update(...)


hynekcer:
The method [bulk_update](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#s-bulk-update) is new in Django 2.2.

Implement it in Django-salesforce.

Notes: A method `salesforce.dbapi.driver.RawConnection.sobject_collections_request(...)` was written for this purpose. The implementation in Django is completely different. Therefore we will not use any code from a Django method and easily backport it for Salesforce backend also to an old Django.

garrypolley:
Is there active work on this issue taking place in a different branch/fork? I'm thinking of potentially looking into helping out here. We have some API limit issues with our updates and batching the updates in groups via the bulk_update would help us out a lot.

hynekcer:
Thanks for your interest and the offer. I can write some useful details: I think that the driver part with `cursor.db.connection.sobject_collections_request('PATCH', records)`  has been worked correctly since v0.8.1. We should (and you can start it until you can) monkey paste the method `bulk_update` ([source code](https://github.com/django/django/blob/efc1c73bf511cbebf03fa5b13fa3a98718daf928/django/db/models/query.py#L517-L559)) to `salesforce.backend.query.SalesforceQuerySet`. You see there a similar `bulk_create`
 The problematic useless part in the original `bulk_update` is with CASE ... WHEN ... is [here](https://github.com/django/django/blob/efc1c73bf511cbebf03fa5b13fa3a98718daf928/django/db/models/query.py#L548-L550). Therefore we will not call it by `super()` at all and useful parts like checks should be copied to our method.

Also tests are useful also for understanding the SFDC API. I have test only for `bulk_create` in `tests.test_mock.test_data.test_create()` (I use "mocked" simplified recorded real responses of new SFDC API features that were documented with no or little examples because sometimes later after an new SFDC version it is easier for maintenance (if a problem is not found while in the release preview window). I think that e. g. error reporting with or without "allOrNone" has been changed in some `/services/data/{API_VERSION}/composite/sobjects` and the change can not be verified because if even doesn't depend on API_VERSION parameter but only the release.)

[...]

PR #252, opened 2020-04-07
Fix "defaulted on create"

### LINKED ISSUE #213: DEFAULTED_ON_CREATE not supported in alternate testing DB?
Hi!
I've successfully set up an alternate testing DB with SQLite as instructed in the readme, however when running my tests, I get the following error:
```
django.db.utils.InterfaceError: Error binding parameter 1 - probably unsupported type
```

This appears to come from a field I have defined as:
```ruby
is_active = models.BooleanField(custom=True,
                                db_column='Is_Active__c', 
                                default=models.DEFAULTED_ON_CREATE)
```

A quick look at the `execute` method in `django/db/backends/sqlite3/base.py` reveals that the `Is_Active__c` param that django is trying to save is:
`<salesforce.backend.operations.DefaultedOnCreate object at 0x10f233c50>` ... instead of True/False. Changing that param value to either True or False while debugging does indeed solve the problem.

Is there a known issue working with DEFAULTED_ON_CREATE in alternate DBs? Am I missing some kind of configuration?

Thanks!

hynekcer:
Fixed #213,  #236 and #250
It will be probably soon a new version.
Removed 2.7 snd 3.4 code and added many static typing annotations in Python 3.5+ style

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ htmlcov/
 coverage_html/
 log
 tests/inspectdb/models.py
+tests/inspectdb/dependent_model/models_template.py
 salesforce/packages/
 venv
 tests/x_*/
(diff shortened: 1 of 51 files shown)


hynekcer:
Closed in favor of PR 255

PR #269, opened 2020-06-18
Authenticate by Refresh Token (replacement of password)

hynekcer:
The last authenticator the completed everything on the refactored modular `auth.py` is the authentication by refresh token, class `salesforce.auth.RefreshTokenAuth`. It is briefly desribed in Wiki [Refresh Token Auth](https://github.com/django-salesforce/django-salesforce/wiki/Experimental-Features#refresh-token-auth). It allows to get a refresh token interactively by web browser by opening one link from console and copying another final link after authorization back to console. The refresh token written on the console should be saved to settings to remain authenticated also after restart.

I prefer to commit it not before v1.1

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -34,8 +34,8 @@ Install, configure a Salesforce connection, create a Salesforce model and run.
 
     'salesforce': {
         'ENGINE': 'salesforce.backend',
-        'CONSUMER_KEY': '',                   # client_id
-        'CONSUMER_SECRET': '',                # client_secret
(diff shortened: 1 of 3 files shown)


philchristensen:
Sounds good. I should be able to get 1.1 released in the next day or two, and then we can merge this.

hynekcer:
Thanks. It does not hurry. I have committed just the version number and I'm preparing for weekend.

philchristensen:
I'm sorry I always say things like this, then take forever to actually do them.

PR #273, opened 2020-08-05
Version 3.1 compatible with Django 3.1 final

hynekcer:
We must release a new version because `salesforce/backend/__init__.py ` supported only Django 3.1 development releases, bun not the 3.1 final. A simple command "pip install django django-salesforce" would not work
It is my mistake that I thought first about some new code for August release or the support should have been added already to the late 1.1.

It is strange that we must release a new version, we have no new code except the fixed version check and we wanted to synchronize the version number in August with 3.1.

Even the unofficial [Refresh Token Authentication](https://github.com/django-salesforce/django-salesforce/wiki/Experimental-Features#refresh-token-auth) is in our v1.1

Nevertheless, I propose to publish 3.1 (or you can rewrite the last commit and CHANGELOG)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,4 +23,4 @@ install:
 script:
   - rstcheck README.rst
   - echo -e "SF_PK = 'Id'" > salesforce/testrunner/local_settings.py
-  - tox -r -e py36-dj30,py36-dj22,py35-dj111,debug_toolbar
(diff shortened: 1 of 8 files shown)


philchristensen:
I like the idea of doing a sort of maintenance release to support django 3.1 and make the versioning change official.

PR #278, opened 2020-11-22
Fix inspectdb

hynekcer:
The first part is almost the same as the previous closed PR #276 Fix275, only with some fixed grammar typos and that Winter'21 is now available everywhere.

The second part is a series of fixes in order to can produce smaller `models.py` by `inspectdb`. A review of this part is desirable. I'm sure that the current solution is not perfect, but better than the previous state. 

A test passed by CI is not important enough. A review of better readability and usefulness of simplified output models.py is also important.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -48,8 +48,9 @@ Install, configure a Salesforce connection, create a Salesforce model and run.
      the Salesforce REST API Documentation. Key and secret can be created on
      web by:
 
-     - Salesforce web > Setup > App Setup > Create > Apps > Connected apps >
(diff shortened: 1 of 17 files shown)


philchristensen:
So would we release this as 3.1.1?

hynekcer:
Please pull it to a new branch.
The second part still requires a documentation and a change log note. I'm going to try the simplified inspectdb in some new project how is it likeable this week and then probably to prepare all for a release. In the worst case the release 3.1.1 will be created from the first part.

Issue #221, opened 2019-01-31
Helping with Django-Salesforce development


ndingdw:
Hi guys!
 
We have been using your project in our Django App for more than two years and we are currently migrating our codebase to Python3.7 and Django2.2 LTS. Noticed that the current release doesn't support the values(), values_list(), defer(), only() and some raw() queries, we are wondering that could you give us the access of the development repository? And  we can help with the testing part!
 
Thank you!

philchristensen:
I’m afraid don’t understand. This *is* the development repository and pull requests are encouraged.

Thanks for your support!

nding1718:
Sorry I didn't make myself clear. I found this in the README: 
"Unsupported methods are currently: values(), values_list(), defer(), only() and some raw() queries. (All fixed in a development repository, waiting for review, consensus etc.)"
So the current repository we have and this fixed "development repository " are the same one?

Thank you.

philchristensen:
I guess I’m not what other repository you’re referring to. This is the only repo there is or has ever been.

nding1718:
OK, I see. really appreciate the quick response. Thank you very much!

hynekcer:
Yes, I promised to release a new version by end of January. I'm working on a big refactoring and because I do it alone now I try to reorganize the last unpublished commit history finally to clean readable commits by "git rebase --interactive", you are right and maybe I commit all for the new version today here for testing.

I worked in 2016 on a similar big principal refactoring, but the difference was so big that it was not possible to merge or combine the branches. Every old idea was now assessed and implemented again or blind alleys were rejected. It cost a lot of time. I implemented some operations by SOAP API, where I even considered to accept reject to be maintainer of Beatbox package, because REST API did not support any for of fast operation on collection of objects. ([SObjects collections](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_sobjects_collections.htm) exist since Salesforce 42.0). I discovered a small security issue in some form in many mainstream databases by work on query compiler for django-salesforce, that I verified in C code in some of them and reported to these projects etc...

nding1718:
That sounds great! And also where can we find the latest developing repository? Or do you have a task list? we can help with them!
Thank you!

hynekcer:
[...]

Issue #172, opened 2016-12-12
beatbox: PythonClient does not exist in beatbox3 only Client


aventurella:
https://github.com/django-salesforce/django-salesforce/blob/master/salesforce/utils.py#L34-L35

In `salesforce.utils. get_soap_client` the client is accessed via `PythonClient`. However the Python 3 port of beatbox (https://github.com/superfell/Beatbox) does not contain `PythonClient` only `Client`.

In my unit tests I was able to this and it seemingly works:

```
from salesforce.utils import convert_lead
import beatbox
beatbox.PythonClient = beatbox.Client
```

I wonder if this couldn't be refactored to accommodate it:

```
import sys

if client_class is None:
        if sys.version_info[0] == 3:
            client_class = beatbox.Client  # https://github.com/superfell/Beatbox
       else:
            client_class = beatbox.PythonClient
```

hynekcer:
I am writing a long answer

aventurella:
@hynekcer you are literally the most responsive project maintainer I have ever encountered. Well done! and Thank you for all you do!

hynekcer:
I'm glad that you wrote this issue. I knew about it, I hoped that this issue will help me to negotiate a related problem and I'm surprised that it came up late.

The reason is a stalled communication about an improved package Beatbox: "[Any new release / pypi release scheduled for Beatbox?](https://github.com/superfell/Beatbox/issues/43)". A made two versions of updated Beatbox compatible with Python3 for two main Beatbox repositories that I had expected one of them will be preferred by feedback of other people and merged and uploaded to PyPI soon. I'm not able to judge the problem that the original name Beatbox was de -facto temporary overwritten by a fork and the original author don't naturally like to be resposible for extraneous extensions and the last maintainer of the fork don't like any backward compatible changes be experienced by current users. The PyPI package Beatbox references to Superfell repository which had not has not been its direct source for many years.

I currently use a dependency on a temporary version  in [django-salesforce tests](https://github.com/django-salesforce/django-salesforce/blob/master/tox.ini#L35):  
`git+https://github.com/hynekcer/beatbox-davisagli.git@7f628a789cba#egg=beatbox`  
more compatible with [davisagli](https://github.com/davisagli/beatbox) (owner of a PyPI package Beatbox). The word "temporary" means that I don't want to arbitrate them or push. I reference a fixed commit be be clear that I don't want to add a new Beatbox code until a basic direction in Beatbox will be chosen and we can incline to it.

[...]

PR #290, opened 2021-07-16
Migrations

hynekcer:
The most interesting new feature is support of **migrations on salesforce.com database** cloud. [Documentation for migrations](https://github.com/django-salesforce/django-salesforce/blob/migrations/docs/migrations.rst)

This pull request and the branch [django-salesforce/migrations](https://github.com/django-salesforce/django-salesforce/tree/migrations) is useful development on sandboxes. It will be not merged into master until migrations are good enough for production.

### Changelog
* Add: Written documentation docs/migrations.rst about the system how to
  restrict objects that should be managed in SFDC, which instance can be
  managed and which custom models can be never destructed e.g. in production
  instance even that they cen be destructed and created again in a development
  instance.
* Add: Command "migrate" can create, update, rename and delete models and fields
  in Salesforce databases.
* Add: Introspection use metadata for better detection of default values
  in inspectdb.
* Fix: Makemigrations with custom models
* Add: Recognize standard fields in sustom models
* Add: Support timestamps with "auto_now_add=True" and "auto_now=True".
* Fix: Fix tests for Salesforce API 52.0 Summer '21 that broke the evaluation of
  .filter(field__gt=''). Method .exlude(field=None) must be used.
* Fix: Works also with obsoleted USE_TZ=False #221
* Fix: Support also alternative clones of Beatbox #172


hynekcer:
Written Quick Srart to docs/migrations. Rewritten a big part of documentation.

PR #294, opened 2021-10-11
Django 4.0 alpha 1

hynekcer:
* Add suport for Django 4.0 alpha 1 and declare support for Python 3.10 final,
  Salesforce API 53.0 Winter '22.
* Add: Support timestamps with "auto_now_add=True" and "auto_now=True".
* Fix: Fix tests for Salesforce API 52.0 Summer '21 that broke syntax of
  filters on a primary key or foreign keys: can not be compared to empty string
  and allowed only =, !=, IN, NOT IN.
  A ``.filter(field__gt='')`` must be replaced e.g. by ``.exlude(field=None)``.
* Fix: Works also with obsoleted USE_TZ=False #221
* Fix: Support also alternative clones of Beatbox #172

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,15 @@ experimental.
 
 [3.2.1] Unpublished
 -------------------
+* Add suport for Django 4.0 alpha 1 and declare support for Python 3.10 final,
+  Salesforce API 53.0 Winter '22.
+* Add: Support timestamps with "auto_now_add=True" and "auto_now=True".
+* Fix: Fix tests for Salesforce API 52.0 Summer '21 that broke syntax of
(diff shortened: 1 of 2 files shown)


hynekcer:
Not now. Tests integration should be updated first. I read this on Travis:
> Since June 15th, 2021, the building on travis-ci.org is ceased. Please use travis-ci.com from now on.

I have only rest credits in free 30 days trial on travis-ci.com that will expire at the end of October. Do you use Travis for some other purpose? Otherwise it is overkill.

philchristensen:
It seems like these days we ought to be able to
move all that to GitHub Actions. I’d love to take a whack at that now that it’s more mature.

PR #295, opened 2021-10-11
Travis update

hynekcer:
If you want to continue with Travis then please set "Limit concurrent jobs" to 1 and test this PR that supports new Python versions and is also faster than compiling Python from source by pyenv. Otherwise this is useless.


philchristensen:
Yeah, I’m going to start by merging this and hopefully it will just work on travis-ci.com. Then I’ll separately work on the GitHub Actions migrations.

PR #296, opened 2021-10-11
WIP: Another attempt at implementing GitHub Actions

philchristensen:
Since Travis-ci.com has started applying complicated pricing rules, it's as good a time as any to look at bringing this "in-house" so to speak with GitHub actions.


philchristensen:
I just need to figure out the best way to get the Python environment ready for tox. This will be worth the hassle.

PR #306, opened 2022-03-08
Fix #302

### LINKED ISSUE #302: Querying ContentDocumentLink Objects.
Using Version django-salesforce 4.0. 

I have the following model:

```
class ContentDocumentLink(models.Model):
    linkedentityid = models.CharField(db_column='LinkedEntityId', max_length=18)
    contentdocumentid = models.CharField(db_column='ContentDocumentId', max_length=18)
    isdeleted = models.BooleanField(db_column='IsDeleted')
    systemmodstamp = models.DateTimeField(db_column='SystemModstamp')
    sharetype = models.CharField(db_column='ShareType', max_length=40, blank=True, null=True)
    visibility = models.CharField(db_column='Visibility', max_length=40, blank=True, null=True)

    class Meta:
        managed = False
        db_table = 'ContentDocumentLink'
```

Do to some limitation on salesforce we cannot do:
```
ContentDocumentLink.objects.all()
```
because it generates the query:

```
SELECT ContentDocumentLink.Id, ContentDocumentLink.LinkedEntityId,
    ContentDocumentLink.ContentDocumentId, ContentDocumentLink.IsDeleted,
    ContentDocumentLink.SystemModstamp, ContentDocumentLink.ShareType, ContentDocumentLink.Visibility
FROM ContentDocumentLink
```

that will throw the error:
```
MALFORMED_QUERY: Implementation restriction: ContentDocumentLink requires a filter
    by a single Id on ContentDocumentId or LinkedEntityId using the equals operator
    or multiple Id's using the IN operator.
```

When I try to do:
```
ContentDocumentLink.objects.filter(contentdocumentid__in=['0693X00000IJWDNQA5', '0693X00000IJWDOQA5']).
```
the generated query:
```
SELECT ContentDocumentLink.Id, ContentDocumentLink.LinkedEntityId,
    ContentDocumentLink.ContentDocumentId, ContentDocumentLink.IsDeleted,
    ContentDocumentLink.SystemModstamp, ContentDocumentLink.ShareType, ContentDocumentLink.Visibility
FROM ContentDocumentLink
WHERE
    ContentDocumentLink.ContentDocumentId IN ('0693X00000IJWDNQA5', '0693X00000IJWDOQA5')
```
throws the same error because salesforce does not recognise "ContentDocumentLink.ContentDocumentId" as being the same as "ContentDocumentId", i.e., one of the mandatory fields.

I've been able to execute the following raw query without problems:

```
ContentDocumentLink.objects.raw(
    """SELECT ContentDocumentLink.Id, ContentDocumentLink.LinkedEntityId, ContentDocumentLink.ContentDocumentId,
ContentDocumentLink.IsDeleted, ContentDocumentLink.SystemModstamp, ContentDocumentLink.ShareType,
ContentDocumentLink.Visibility
FROM ContentDocumentLink WHERE ContentDocumentId IN ('0693X00000IJWDNQA5', '0693X00000IJWDOQA5')""")
```

My conclusion is that salesforce assumes that the ContentDocumentId (or LinkedEntityId) must be present on the where part of the query but does not accept ContentDocumentLink.ContentDocumentId.

Is it possible to force the django generated sql queries to drop all the explicit table names of the fields? Maybe only on the where part of the queries?
Is there a better solution?

hynekcer:
A simple unit test for the issue is written. Maybe @gredondogc prepares some simple test model and data and I write an integration test with that data.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,8 +14,15 @@ Some items here can be marked as "internal": not ready enough or
 experimental.
 
 
-[4.0] Unpublished
------------------
+[4.0.1] Unpublished
+-------------------
+* Fix: Optionally don't use redundant table names before field names
+  if queried with minimal_aliases=True; important for some special system objects #302
+* Fix: Tests with the newest Django, Salesforce, Python; including Python 3.11 (alpha)
+* Fix: Extended SalesforceModel with PostgreSQL backend and Django >= 3.0 #299
+
+[4.0] 2021-11-22
+----------------
 * Internal change: The default row type from salesforce Cursor is now a tuple,
   not a list
 * Fix: Invalid primary key from bulk_create([one_object]) in Django 3.0 #298
--- a/salesforce/backend/compiler.py
+++ b/salesforce/backend/compiler.py
@@ -292,7 +292,7 @@ class SQLCompiler(sql_compiler.SQLCompiler):
         if self.soql_trans is not None:
             return self.soql_trans
         if not _alias_map_items and not self.query.alias_map:
-            # empty alias_map is possible due to field expr in Django 1.8
+            # empty alias_map is possible due to field expr
             return {}
         # Unified interface:
         #   alias_map_items = [(lhs, table, join_cols_, rhs),...]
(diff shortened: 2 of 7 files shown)


hynekcer:
We can wait with the new release 4.0.1 a few days until a test for this issue 302 is added and maybe something small related could need to be fixed.

hynekcer:
That everything was a part of release 4.1, but separated to more commits related or similar.

Issue #299, opened 2021-12-03
Django `SystemCheck` fails on KeyError `CharField` with PostgreSQL


piotrpawlaczek:
I run into the following issue when upgrading to `django3.2`

django-salesforce: `4.0` and `3.2`
django: `3.2.9`

Reproducible with cmd:
```shell
>> ./manage.py check
```

Skippable with cmd:
```shell
>> ./manage.py shell_plus --skip-checks
```

Traceback
```python
>> ./manage.py check

Traceback (most recent call last):
  File "src/manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/virtualenv/lib/python3.8/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/virtualenv/lib/python3.8/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/virtualenv/lib/python3.8/site-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/virtualenv/lib/python3.8/site-packages/django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
  File "/virtualenv/lib/python3.8/site-packages/django/core/management/commands/check.py", line 63, in handle
    self.check(
  File "/virtualenv/lib/python3.8/site-packages/django/core/management/base.py", line 419, in check
    all_issues = checks.run_checks(
  File "/virtualenv/lib/python3.8/site-packages/django/core/checks/registry.py", line 76, in run_checks
    new_errors = check(app_configs=app_configs, databases=databases)
  File "/virtualenv/lib/python3.8/site-packages/django/core/checks/model_checks.py", line 34, in check_all_models
    errors.extend(model.check(**kwargs))
  File "/virtualenv/lib/python3.8/site-packages/django/db/models/base.py", line 1271, in check
    *cls._check_fields(**kwargs),
  File "/virtualenv/lib/python3.8/site-packages/django/db/models/base.py", line 1382, in _check_fields
    errors.extend(field.check(**kwargs))
  File "/virtualenv/lib/python3.8/site-packages/django/db/models/fields/__init__.py", line 2459, in check
    *super().check(**kwargs),
  File "/virtualenv/lib/python3.8/site-packages/django/db/models/fields/__init__.py", line 1773, in check
    *super().check(**kwargs),
[...]

Issue #280, opened 2020-12-03
DefaultedOnCreate object Is the default value instead of Salesforce value


noamby:
Hi :)

I have an issue when creating a model with field using DefaultedOnCreate()
The field is not updated with the actual value in Salesforce, it's value would be DefaultedOnCreate object

Example Model:
```
class Contact(SalesforceModel):
  currency_iso_code = TestableCharField(db_column='CurrencyIsoCode', max_length=3, verbose_name='Contact Currency', default=models.DefaultedOnCreate('USD'), choices=[('USD', 'U.S. Dollar')], blank=True, null=True)
  last_name = models.CharField(max_length=80)
```

Example Code:
```
contact = Contact.objects.create(last_name="test contact")
contact.currency_iso_code
# will print <salesforce.backend.operations.DefaultedOnCreate object at 0x10f543cf8>
```

hynekcer:
Yes, it is currently an expected feature of Salesforce that can be circumvented in several possible ways by you.  A simple information about it should be added to README. Too complicated information can be found in our [wiki about default=models.DEFAULTED_ON_CREATE](https://github.com/django-salesforce/django-salesforce/wiki/Introspection-and-Special-Attributes-of-Fields) A "fix" is also possible, but it would require a discussion, because a fix could be undesirable for some people. I explain the current state first.

Salesforce cloud has a feature that a very sophisticated method can be be used to determine a default value if a field is omitted in a new object that is created. Your default currency can be e.g. determined by settings of the organization and you can use the same code for many organizations that have different default currencies. If your case is not so complicated then you can simply replace it by `default='USD'`. A typical example and maybe the only example where a DefaultedOnCreate is necessary almost for everyone is the "Owner" field of the object that is can be a current user in a simplest possible case but not `None`.

Another solution is to refresh the value from the database: `contact = Contact.objects.get(pk=contact.pk)` or some [refresh function](https://github.com/django-salesforce/django-salesforce/blob/79ba20d6b03a88e72581e71917a3cc636ac21391/tests/test_mixin/tests.py#L6-L11) if you really need a more dynamic default assigned by Salesforce, not by Django.

A parameter in `models.DefaultedOnCreate('USD')` is intended only for possible tests of your application with SalesforceModel with other databases that have no DefauledOnCreate feature and it is ignored with the Salesforce database.

(Recently it is also possible to "fix" that issue in this framework by using a composite request instead of simple REST request to get also a new value of your field by same request that created the object.  
[...]

PR #309, opened 2022-07-26
Version v4.1 for Django 4.1

hynekcer:
Django 4.1rc1 has been released on July 19. We can release django-salesforce v4.1 now.

* Add: Support for Django 4.1
* Add: Command ``inspectdb`` can introspect actual default values
  of fields from a ``defaultValueFormula`` if it is a simple constant
  like a number or a string.
* Fix: A default value ``DefaultedOnCreate(value)`` is no longer created
  by ``inspectdb`` in favour of a simple ``value``. If a simple default value
  can not be known then a generic ``DEFAULTED_ON_CREATE`` is still used rarely
  for default values created by a complicated or unknown function only
  on Salesforce side. #280
* Fix: Optionally don't use redundant table names before field names
  if queried with ``.sf(minimal_aliases=True)``; important for some
  special system objects #302
* Fix: Tests with the newest Django, Salesforce, Python; including Python 3.11(beta)
* Fix: Extended SalesforceModel with PostgreSQL backend and Django >= 3.0 #299

A simple release draft is also prepared (the same text)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,4 +26,4 @@ jobs:
         - python3.8 -V
       env: TOXENV=docs_style,typing,dj20-py36,dj22-py38,dj30-py36,dj32-py38
     - python: 3.9
-      env: TOXENV=dj40-py39,no_django-py39,debug_toolbar-dj32-py39
+      env: TOXENV=dj41-py39,dj40-py39,no_django-py39,debug_toolbar-dj32-py39
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,8 +14,8 @@ Some items here can be marked as "internal": not ready enough or
 experimental.
 
 
-[4.0] Unpublished
------------------
+[4.0] 2021-11-22
+----------------
 * Internal change: The default row type from salesforce Cursor is now a tuple,
   not a list
 * Fix: Invalid primary key from bulk_create([one_object]) in Django 3.0 #298
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,24 @@ Some items here can be marked as "internal": not ready enough or
 experimental.
 
 
+[4.1] Unpublished
+-----------------
+* Add: Support for Django 4.1
+* Add: Command ``inspectdb`` can introspect actual default values
+  of fields from a ``defaultValueFormula`` if it is a simple constant
+  like a number or a string.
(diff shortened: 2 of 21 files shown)


philchristensen:
I'm not sure what's causing these Travis issues, but I'm determined to get them fixed ASAP.

hynekcer:
The first problem is that it runs without environment variables: 'CONSUMER_KEY', 'USER', 'PASSWORD', 'CONSUMER_SECRET'. The second problem is that the new flake8 version 5.0 (5.0.0 is five days old and today is a fix 5.0.4) uses different rules to find configuration files. I updated the configuration also in setup.cfg in the last commit. Also the previous origin/main would fail with with new flake8 after `tox --recreate` in my test environment.

> This seems to be caused by https://github.com/PyCQA/flake8/pull/1472, previously all setup.cfg", "tox.ini", ".flake8" where considered however the code now only considers the first file it finds (in this order) with a flak8 section. [Flake8#1639](https://github.com/PyCQA/flake8/issues/1639)

hynekcer:
Fixed configuration for flake8. Test passed in the main branch easily.

Django 4.1 has been released yesterday and we can follow them.

philchristensen:
Yeah, not sure what happened to those environment variables, but I'm looking into it now.

philchristensen:
huh, now it looks like the jobs passed, but I still see Xs above.

philchristensen:
Oh, i see you fixed it....still trying to understand what happened...

philchristensen:
Well, I don't need to understand it to do the release ;-) Just a moment...

philchristensen:
https://pypi.org/project/django-salesforce/4.1/

hynekcer:
Thank you.

It is probably a security policy that your private tokens should be never not passed to pull requests from unknown person. The most restrictive policy is to pass environment variables only to a code in the main branch. And if I remember, something between can be configurable. (I had only 30 day trial on Travis last year with a few hundred free minutes that were consumed soon.)

philchristensen:
Ohhhh, you're right, I remember now. This has always been an issue. Usually I pull your changes and re-push them as a separate branch on the main django-salesforce repo, which re-triggers the job from there, leading to the correctly decrypted env vars.

That's why your second `setup.cfg` update worked, you made it on the `main` branch. All good.

I really need to take another whack at automating the release process. Previously I got wrapped up in the idea of re-doing the whole testing process as a Github Action, but it would be enough to just create a manual release job that we could run as needed.

PR #311, opened 2022-08-05
Travis jammy

hynekcer:
* Python 3.6 is after end of life. A new Ubuntu 22.04 LTS Jammy Jellyfish is an option on Travis.
* I test frequently also with Python 3.11 beta in tox locally.


philchristensen:
FYI, you can test in travis yourself if you just push your development branch to this repo, since you're an admin.

If you follow your normal workflow, you can just add another remote that points here, and push it after you make the PR. The UI is smart enough to just look at the most recent build.

An alternative would be pushing your branch here before creating the PR from that directly.

hynekcer:
Thanks. I tried it. That is very useful.

Issue #315, opened 2022-10-27
Suppport for Django cache framework with caching in other database 


hynekcer:
Django can store its cached data by different backends like Memcached, Redis... or in a database for caching of views, parsed template fragments or any cached serializable objects. A local 'default' database is much faster than a view combined from data from more Salesforce requests. That implies to use two databases. (`salesforce` and `default`) Django Database caching requires to write a router class that determines which database should serve the cache table if multiple databases are used. (described in Django docs [Database caching](https://docs.djangoproject.com/en/4.1/topics/cache/#database-caching))

That code is now added as a part of the standard `salesforce.router.ModelRouter`.

hynekcer:
Implemented - test passed

PR #316, opened 2022-10-27
Support Django DatabaseCache with 'default' database by ModelRouter #315

hynekcer:
Django caching to a 'default' database can be now used without adding a database router that would be otherwise necessary in a case of multiple databases. This is the usual way how tables for models with SalesforceModel are in `salesforce` database and other possible tables are in `default` database including a possible django cache table that has no explicit model or application


hynekcer:
(Added an unimportant commit to restart a [test that passed](https://app.travis-ci.com/github/django-salesforce/django-salesforce/builds/257165068) with environment variables.

Issue #320, opened 2023-05-10
Compatibility with Django 4.2


joshuadavidthomas:
Django 4.2 was released just over a month ago. This package has a check in `salesforce/backend/__init__.py` that prevents using it with the newest version. (There is no enforcement of the version constraint in the package's `setup.py`.)

4.2 introduced some breaking changes for the database backend API -- https://docs.djangoproject.com/en/4.2/releases/4.2/#database-backend-api -- that may need to be dealt with.

philchristensen:
There’s no specific timeline for release, as this project has a very small support team.

Users with production timelines impacted by this are encouraged to leverage the GitLab sponsorships system, which directly funds the primary developer of this project. You may also wish to DM @hynekcer directly.

joshuadavidthomas:
Thanks for the quick response! Are the maintainers open to an outside PR to tackle this? If so I may take a crack at it, time permitting.

Thanks to all the contributors and maintainers for this package! 😄

philchristensen:
Oh, PRs are encouraged and appreciated!

hynekcer:
I prepared a fix for Django 4.2 on March 15th. The problem is that this project has very few sponsors and little total contribution, as you can see on the page https://github.com/sponsors/hynekcer.

I am thinking about incentives for sponsors. Words alone did not help enough. My idea is to continue this project in two very similar repositories under two licenses. One repository with unrestricted code under very a restrictive license AGPL (GNU Affero General Public License - which  requires to share every code that uses it including e.g. any web application) and this repository will continue unchanged under MIT license, but some new code in this repository added after April 2023 would call a function "sponsor_motivation()" that will pass silently if you have a sponsor code in your settings or it rejects to continue.

The restricted code should be usually support for the newest stable Django version (development Django versions remain unrestricted) and Django migrations. I have been using migrations for two years and I think that it is a very useful part of a development with Django, but it will require sponsors so that it is reasonable to solve possible secondary problems.

Please [write me](mailto:hynek@sdb.cz) what should be a part of your sponsor code (a company name and a contact email to you).

hynekcer:
Version 4.2 that is compatible with Django 4.2 has been released yesterday. See the README about license alternative or ask for a license key.

Issue #312, opened 2022-08-17
SF picklist fields with lots of choices not being fetched on inspectdb


lsimps204:
I'm having a problem while upgrading an old application using an early version of `django-salesforce` to the most recent version.

Many fields in the app fetch choices from Salesforce picklists, and display them on the UI - however, these choices appear to be truncated in the newer versions after running `inspectdb`, with this comment following the field: `# Too long choices skipped`

Is there any way around this? I've had a look at the source but I don't see any alternative that would allow the choices to be fetched, but I need to display these fields and allow users to select a valid choice.

hynekcer:
The length of picklist was restricted to 4000 characters in django-salesforce 3.2 (April 2021) because so long lines can not be edited. The idea is that two typical long picklists are countries and list of time zones. These picklist are repeated in many objects, but they should be probably shared, if they are used. The most terrible picklist is `time_zone_sid_key` approximately 33000 charactes long which is changed many times in year because counties change the Daylight Saving time on different days in Spring and Autumn.

I can make the limit configurable by settings variable `SF_MAX_INSPECTDB_PICKLIST_LENGTH`.

hynekcer:
Fixed in the main branch. I rename the issue because the word "Upgrading" is not important for other people for the most recent upgrade. I accept that this was an improvement that is also a regression.

lsimps204:
Many thanks for the quick action! I understand the rationale behind the 3.2 changes, but it's good to have the ability to configure the limit - thanks again!

PR #321, opened 2023-05-23
Django 4.2

hynekcer:
- Support for Django 4.2  #320
- check is_enterprise_key()
- Use API 57.0 Spring'24
- Suppport for Django cache framework with caching in other database #315
- Configurable max pick-list size "SF_MAX_INSPECTDB_PICKLIST_LENGTH" #312

The code is here in the branch django24.
Some texts should be revised before release 4.2, but all code is ready.


salesforce/backend/enterprise.py
```diff
@@ -0,0 +1,46 @@
...
+
+    # This check is an integer arithmetic puzzle that does not use anything except Python
+    # and Django settings. It is evaluated in less than half milisecond on a notebook CPU.
+    # There is no cryptography, no network access. The license key never expires.
+    # It is a safe code. Please do not discuss publicly about other aspects how the
+    # license key is weak or strong. Many easier ways exist how to skip a verification
+    # without analyzing this. :-)
+
+    # A legal way to skip the enterprise check is to use django-salesforce-agpl and
+    # accept the restrictive AGPL licence that requires you provide all your Django
```
sebastian-philipp, line 27: Note that the license.txt in the repo root still uses the MIT license.

philchristensen: Seems like forking and removing the protection wouldn't be allowed by GPL either. I don't know if I understand what the goal is of the license change.

My attitude is "don't break into my house because I asked you not to, not because there are bars on the windows"

hynekcer: Exactly. Open doors cannot be kicked out. That is why I initially thought about name "sponsor_motivation()", not "...key". I think a some short permissive license is fine with a short restriction related to the enterprise-key code that should be not removed or evaded because it is important for the project.
The AGPL license however (as a copyleft license) also doesn't protect the author's work from copying, modifying, sharing. It establishes a non-dispensable mandatory freedom that everybody who uses that code must share with others and publish also all his/her code under the same copyleft license and give other users the same right on his work, even to end users who do not need a software and only use a service by the network. "Network use is distribution" is a principle of AGPL. Copyleft makes a strong border between "free" open-source world and commercial world. People should not remove the enterprise-key code, but they can use, modify, share... a version under AGPL where the enterprise check is removed by us.

hynekcer:
I modified the MIT license so that it is still recognized as MIT. It is a hack witch `<br>` tags in the first line where is a place for "Copyright Year Authors" in MIT license.
Please fix it or comment it.

The same example is now in https://github.com/hynekcer/django-salesforce  
where it is used now in the main branch, because the icon "License" on Github in the right side-bar is synchronized only with the main branch.

philchristensen:
Okay, I get it now. This all looks good to me now.

philchristensen:
oh, i'll add that key to travis on this branch too...

hynekcer:
I prepared tests by Github actions. Some bugs must be still fixed for that environment.

hynekcer:
Fixed and tests [passed](https://github.com/django-salesforce/django-salesforce/actions/workflows/main.yml) on Github.

hynekcer:
Pushed because it is definitely better than a previous code without Django 4.2. Test badge could not be updated without pushing "actions" code into main branch.

PR #323, opened 2023-05-25
Readme: test result badge by Github

hynekcer:
-  Readme: Add test result badge by Github to home page (it is visible now in the branch "docs")

What to do with Travis? It was better that Github Actions, but...

Some readme will be still beneficial.


philchristensen:
While I do agree we should move away from Travis, I'll still reach out to them today and get the credits fixed. I had asked last time if it could be done automatically but that wasn't an option.

PR #339, opened 2024-08-28
Limit package compatibility to what the package actually supports

### LINKED ISSUE #338: Limit Django compatibility on package level
Currently the package defines itself as compatible with `django>=2.1` in [setup.py ](https://github.com/django-salesforce/django-salesforce/blob/main/setup.py). However on runtime, the [initialization](https://github.com/django-salesforce/django-salesforce/blob/main/salesforce/backend/__init__.py) makes additional checks that prevent it from working with Django 5.1.

This should ideally be prevented and the package should already report to be incompatible with 5.1  
Otherwise, this might break installations that upgrade Django in good faith that nothing is blocking and then break when actually running them.

Thanks!

lociii:
This will prevent pip (and similar tools) from allowing to upgrade `Django` to `5.1` as long as `django-salesforce` does not support it.  
So far, this ended in a runtime issue which might break things.

Fixes #338

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,7 @@ def autosetup():
         # setuptools won't auto-detect Git managed files without this
         setup_requires=[] if not with_git else ["setuptools_git >= 0.4.2"],
 
-        install_requires=['django>=2.1'] + requirements_txt,
+        install_requires=['django>=2.1,<5.1'] + requirements_txt,
 
         # metadata for upload to PyPI
         author="Hynek Cernoch",


hynekcer:
A simple fix could be to change `max_django` in `salesforce/backend/__init__.py` and the best fix is the release 5.1
PR #340

lociii:
This is not actually fixing the issue but just moving it to the future. The same problem will appear once 5.2 gets released.  
I would really recommend setting an upper boundary in `setup.py` to mark the package as incompatible with non-supported Django releases. This limit should then be adjusted once `django-salesforce` is officially supporting that Django release.

hynekcer:
I reopened the original issue and explained there why my fix must be different.

PR #341, opened 2024-10-06
Version 5.1

hynekcer:
CHANGELOG

* Add: Support for Django 5.1
* Change: The cursor rollback() method log messages are silenced by default
  because they are mostly anoying in tests. They can be enabled again if
  ``settings.DATABASES['salesforce']['OPTIONS']['WARNING_ON_ROLLBACK'] == True``
* Changes where users will not know a difference:
  * Refactored the backend.query to use consistent data types internally.
  * Checked a development version of Django is really used by "tox -e djdev"
  * Fix introspection for Salesforce API 62.0 Winter '25
    (skip introspection of ...Event object that is not a table)


hynekcer:
I rewrote configuration files for distribution of the package because the new "setuptools" wrote a huge warning "Please avoid running 'setup.py' directly" when I tried to prepare distribution according to our last documented [Release Process](https://github.com/django-salesforce/django-salesforce/wiki/Release-Process). The message links to a blog [Why you shouldn't invoke setup.py directly](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html) where the new command "python -m build" with [build package](https://pypa-build.readthedocs.io/en/stable/) is explained. That is also recommended in [Python Packaging Tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives).

I simplified "setup.py" and rewrote all its static metadata to "pyproject.toml". Then I successfully built the distribution files and pushed it to [test.PyPI/test-django-salesforce](https://test.pypi.org/project/test-django-salesforce/) (tested  with a modified name because I do not have access to django-salesforce). Then I updated the Release Process wiki. (The old text is crossed out, not deleted)

The AGPL version and Release draft v5.1 are also ready. If it is acceptable a please release a new version.

philchristensen:
I had *just* run into the same issue myself only a couple days ago.

This looks good to me, I'll do the release now.

PR #342, opened 2024-10-31
Fix for RemoteDisconnected issue in Login

hynekcer:
This is a hotfix for a reported network issue until that error can be reproduced:

Error 401 after some inactivity and after a message
"RemoteDisconnected: Remote end closed connection without response"

The final fix can be different when more information about the issue is found.


hynekcer:
The fix is merged because the same issue was reported by another user:
```
ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
```

That issue can not be easily reproduced. It was not reasonable to wait so long for a feedback from the original user.

The issue is probably related to Salesforce release Winter '25 or to migration to Salesforce Hyperforce infrastructure architecture.

weinbergdavid:
Thank you; that resolved my issue.

